PR #1, opened 2018-03-05
typos in Univalence from Scratch

pdorrell:
Hi, I'm pretty new to Univalence, and most of my experience of MLTT consists of using Idris as a theorem prover for various toy examples. 

Possibly the variable **y** of type **X** could have a different name, eg **x'** or **x_1**, to avoid confusion with **y** of type **Y** ?


martinescardo:
Thank you.

PR #3, opened 2018-06-06
Added modification of naive-funext-to-funext

cmknapp:
The proof that naive function extensionality implies full function extensionality goes via the fact that if f is an equivalence, then so is precomposition with f. I've abstracted that argument here.

This means that the following are logically equivalent:

1. Naive function extensionality;
2. Precomposition with an equivalence is an equivalence;
3. Function extensionality;
4. Propositions are closed under Pi,
5. Singletons are closed under Pi.

The last 4 are propositions, so are equivalent. Should I add a file summarizing this?


martinescardo:
Feel free to add a file.

PR #6, opened 2019-05-22
The lifting of a set is a set

tomdjong:
Also proved that the product of two sets is a set (see UF-Subsingletons).


martinescardo:
Thanks. Another proof is possible using that `is-prop (l ⊑ m)` for the lifting of a set, proved in LiftingUnivalentPrecategory, and that `⊑` is antiysymmetric, so that we get a constat map `l = m -> l = m` and we can apply Hedberg to get that the lifting is a set. But your proof is also short enough.

martinescardo:
Please import your file at PartialElements.lagda so that `agda AllModulesIndex.lagda` processes it.

tomdjong:
Cool, I like that.

New pull request coming up.

martinescardo:
We should eventually set up an automatic script to compile everything when something is added...

martinescardo:
Please observe the convention of writing e.g. "(X : 𝓤  ̇ )" with a space between the dot and the bracket. This is needed to make the automatically generated html readable.

tomdjong:
Apologies, fixed in new pull request.

martinescardo:
No need to apologize. I need to write a file with convention guidelines.

PR #10, opened 2019-06-27
Update CompactTypes.lagda

tnttodda:


martinescardo:
Please leave a space between the universe dots and the brackets that follow them. This is for correct rendering of the html files generated by Agda.

PR #14, opened 2019-09-05
General and specific classifiers

tomdjong:
Implemented a to-do by Martin in UF-Classifiers.


martinescardo:
Thanks. I also had done this in my lecture notes and the examples, but I hadn't backported it to this repository. Our proofs seem to be different at first sight.

tomdjong:
Ah, it was a bit silly to not check your lecture notes. Your proof certainly seems a lot shorter. Let's properly compare them some time.

martinescardo:
Sure, let's discuss the differences when we meet next time.

PR #17, opened 2019-10-31
Cluster module for PCF and the Scott model

tomdjong:


martinescardo:
Thanks. You should also link this new module from AllModulesIndex.lagda replacing the link to PCF.

PR #39, opened 2020-12-24
Free join-semilattice

tomdjong:
Since last time: I added comments, made the changes you suggested and improved the presentation.


martinescardo (commented):
Shouldn't lines 466-474 be removed?

martinescardo:
Thanks! This is very nice.

Issue #43, opened 2021-03-05
Define combinators for Ω


ayberkt:
It is often convenient to have combinators for Ω, for instance

```agda
_∧_ : Ω 𝓤 → Ω 𝓥 → Ω (𝓤 ⊔ 𝓥)
P ∧ Q = (P holds × Q holds) , γ
 where
  γ = ×-is-prop (holds-is-prop P) (holds-is-prop Q)
```

[agda/cubical](https://github.com/agda/cubical/) has these defined [in this module](https://github.com/agda/cubical/blob/master/Cubical/Functions/Logic.agda). It might be a good idea to do something similar in `TypeTopology`.

ayberkt:
I defined some of these here:

https://github.com/ayberkt/TypeTopology/blob/e23b05dffe61d5aa01436a3378de0bfe63a117fb/source/Frame.lagda#L52-L65

I can send a PR creating a module with these if you are happy with the syntax.

martinescardo:
Yes, please. Thanks.

PR #44, opened 2021-03-08
Add a definition of frames

ayberkt:
I have used combinators for Ω in this development that I defined inside the `Frame` module. Instead of adding such ad-hoc definitions, we should first address #43 in a separate PR in my opinion.

### DIFF
--- a/source/Frame.lagda
+++ b/source/Frame.lagda
@@ -340,3 +340,31 @@ poset-of : frame 𝓤 𝓥 𝓦 → poset 𝓤 𝓥
 poset-of (A , (_≤_ , _ , _ , _ , iss) , p , _) = (A , iss) , _≤_ , p
 
 \end{code}
+
+\section{Frame homomorphisms}
+
+\begin{code}
+
+is-a-frame-homomorphism : (F : frame 𝓤  𝓥  𝓦)
+                          (G : frame 𝓤′ 𝓥′ 𝓦′)
+                        → (∣ F ∣ → ∣ G ∣)
+                        → Ω (𝓤 ⊔ 𝓦 ⁺ ⊔ 𝓤′ ⊔ 𝓥′)
+is-a-frame-homomorphism {𝓦 = 𝓦} F G f = α ∧ β ∧ γ
+ where
+  P = poset-of G
+
+  iss : is-set ∣ G ∣
+  iss = carrier-of-[ P ]-is-set
+
+  open Joins (λ x y → x ≤[ P ] y)
+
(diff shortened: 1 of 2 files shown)


ayberkt:
@martinescardo

Changes we discussed yesterday have now been incorporated and this PR is ready for review.

Removing the explicit h-set requirement has the consequence that my function

```agda
satisfies-frame-laws : {A : 𝓤 ̇} → frame-data 𝓥 𝓦 A → 𝓤 ⊔ 𝓥 ⊔ 𝓦 ⁺ ̇
```

is no longer a proposition. Isn't this problematic? I recall needing this in [`formal-topology-in-UF`](https://github.com/ayberkt/formal-topology-in-UF/).

martinescardo:
This should still be a proposition.

martinescardo:
Nice. Please add: (1) the initial date of creation of the file next to your name at the top of each file, (2) an import entry in theindex file index.lagda (with your name in the imports like the other ones). And (3), make sure "agda AllmodulesIndex.lagda" succeeds. At the moment  I don't have this set up for git to check (but you are welcome to add this feature if you want). Thanks.

ayberkt:
I have made the changes (1), (2), and (3).

> This should still be a proposition.

Antisymmetry is now a structure and not a proposition. It used to be grouped together with the other laws but now that it is not propositional, I would have to group it with the structures to maintain a clean separation of the structure and data involved in defining a frame. Is this what you meant? Otherwise, I don't see how `is-partial` would be propositional. If I already know that a type is equipped with a reflexive and antisymmetric relation, I know that the antisymmetry of the relation is propositional, but this doesn't help me at the point when I'm asserting the antisymmetry.

martinescardo:
> I have made the changes (1), (2), and (3).
> 
> > This should still be a proposition.
> 
> Antisymmetry is now a structure and not a proposition. It used to be grouped together with the other laws but now that it is not propositional, I would have to group it with the structures to maintain a clean separation of the structure and data involved in defining a frame. Is this what you meant? Otherwise, I don't see how `is-partial` would be propositional. If I already know that a type is equipped with a reflexive and antisymmetric relation, I know that the antisymmetry of the relation is propositional, but this doesn't help me at the point when I'm asserting the antisymmetry.

Whether a frame is assumed to be a set, or proved to be a set (from antisymmetry) should *not* make a difference. In the treatment of ordinals, I also don't assume that the underlying types are sets (because this follows from the extensionality condition), but I often do need to know that it is a set in order to prove that certain things are propositions. The same should be the case for you.

PR #47, opened 2021-03-10
Run the Agda action without the `--safe` flag

ayberkt:
This should fix the failing run of the Agda action. The problem is that it defaults to running Agda with the `--safe` flag which doesn't work for `AllModulesIndex.lagda`. This PR disables the `--safe` flag.


martinescardo:
Does this override the `--safe` flag in `index.lagda` and its subtree?

martinescardo:
Excellent. And then we should make our own release that is guaranteed to work with 2.6.1.3 before we move to the next version.

PR #48, opened 2021-03-10
Add combinators for Ω (addresses #43)

### LINKED ISSUE #43: Define combinators for Ω

ayberkt:
_This PR fixes #43._

I have added subsingleton combinators as discussed in #43. Because different combinators require different assumptions, I have separated these into their own submodules. To give an example:

```agda
module Universal (fe : Fun-Ext) where

 ∀[∶]-syntax : (I : 𝓤 ̇) → (I → Ω 𝓥) → Ω (𝓤 ⊔ 𝓥)
 ∀[∶]-syntax I P = ((i : I) → P i holds) , γ
  where
   γ : is-prop ((i : I) → P i holds)
   γ = Π-is-prop fe (holds-is-prop ∘ P)


 ∀[]-syntax : {I : 𝓤 ̇} → (I → Ω 𝓥) → Ω (𝓤 ⊔ 𝓥)
 ∀[]-syntax {I = I} P = ∀[∶]-syntax I P

 infix 3 ∀[∶]-syntax
 infix 3 ∀[]-syntax
 
 syntax ∀[∶]-syntax I (λ i → e) = ∀[ i ∶ I ] e
 syntax ∀[]-syntax    (λ i → e) = ∀[ i ] e
```

It will usually be the case that _all_ of these will be imported, so I also added a submodule called `AllCombinators` (that dependends on both propositional truncation and function extensionality) and imports all of these submodules.

### DIFF
--- a/source/Frame.lagda
+++ b/source/Frame.lagda
@@ -21,7 +21,10 @@ module Frame
        where
 
 open import UF-Subsingletons
+open import UF-Subsingleton-Combinators
 open import UF-Subsingletons-FunExt
+
+open AllCombinators pt fe
 
 \end{code}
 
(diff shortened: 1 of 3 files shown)


martinescardo:
Thanks. All new modules should be added to *index.lagda* unless they are unsafe. We want unsafe modules for limited purposes only (such as showing that type-in-type leads to a contradiction, or working with Brouwerian axioms in a non-standard way by disabling the termination checker (a better way is to work with axioms, given as module or function parameters)).

ayberkt:
>  I wonder how we could check them

My Agda action actually downloads `cubical` every time it runs because it was needed for my development. We could use that to typecheck the modules that depend on `cubical`; this shouldn't be hard to do. However, it would introduce the minor complication of keeping track of the latest release version of `cubical`.

martinescardo:
As I said, the cubical library is a moving target at the moment. If you look at the file mentioned above, you will see some comments explaining what to change depending one whether one uses the released version or the current version under development.
So let's leave this at that for the moment, until the next official release of the Cubical library.

ayberkt:
> If you look at the file mentioned above, you will see some comments explaining what to change depending one whether one uses the released version or the current version under development.

Ah I see what you mean now. Sure, let's leave it like this for now.

PR #52, opened 2021-04-02
Construction of the circle following Bezem et al.

tomdjong:
For review. We can discuss any possible changes in person.


martinescardo:
We have discussed this enough in the Agda Club. Thanks for this contribution!

martinescardo:
html rendering available at https://www.cs.bham.ac.uk/~mhe/TypeTopology/CircleModules.html

PR #63, opened 2021-10-04
Define frame bases and implement their directification

ayberkt:
This PR implements the directification of a frame basis as discussed last Monday.


martinescardo:
Please change "directify-works" to "directify-is-directed".

PR #70, opened 2021-12-23
Formalization of some domain theory 

tomdjong:
* Refactored some existing code on dcpos and the Scott model of PCF in TypeTopology.
* Added the formalization of Scott's D_\inf.


martinescardo (approved):

martinescardo:
Excellent!

martinescardo:
I will update the html version shortly.

martinescardo:
I don't think I am competent enought with git. I don't think your files were committed.

martinescardo:
Why are we getting "closed with unmerged commits"?

martinescardo:
The html is generated now.

tomdjong:
> Why are we getting "closed with unmerged commits"?

Maybe you got confused because approving does not imply merging?

martinescardo:
I think it was because the checks hadn't completed.

ayberkt:
> I think it was because the checks hadn't completed.

If this was accidental, it might be a good idea to enforce the checks by requiring PRs into `master` to pass them before they can be merged. This can be done by going to the branches section of the repository settings and checking the box that says **require status checks to pass before merging**. However, this would prevent the contributors from working directly on the `master` branch.

tomdjong:
I don't think this was the case to be honest. If I recall correctly, I saw that the PR passed the checks before @martinescardo approved it. Indeed, here is the report triggered by the PR: https://github.com/martinescardo/TypeTopology/actions/runs/1616275019. Moreover, the conversation above shows just approving and closing, and only merging after reopening. The reopening then triggered a second check: https://github.com/martinescardo/TypeTopology/actions/runs/1616392461.

Anyway, have a nice Christmas both! :christmas_tree:

PR #74, opened 2022-03-02
Adjoint Functor Theorem and Heyting implications for locally small frames

ayberkt:


ayberkt:
This PR adds three new modules:

1. `GaloisConnection.lagda`,
2. `AdjointFunctorTheoremForFrames.lagda`, and
3. `HeytingImplication.lagda`.

- In `GaloisConnection.lagda`, the definition of adjunction between two posets is given and some elementary facts about adjunctions in the posetal context are proved.
- In `AdjointFunctorTheoremForFrames.lagda`, it is proven that a monotonic map between two frames preserves joins iff it has a right adjoint, assuming that both frames are locally small and the domain has a small basis.
- In `HeytingImplication.lagda`, the Heyting implication is constructed in locally small frames with small bases as a special case of the Adjoint Functor Theorem.

The construction of the Heyting implication will be needed in the definition of the patch frame.

martinescardo (approved):

PR #77, opened 2022-03-15
Small images yield small quotients

tomdjong:
As discussed yesterday and over the weekend.


martinescardo:
Merged - thanks.
But I think it would be better to move the definition of small images to the module UF-Size.

PR #78, opened 2022-03-15
Move definition of small images to UF-Size

tomdjong:
As suggested here: https://github.com/martinescardo/TypeTopology/pull/77#issuecomment-1068240851


martinescardo:
Thanks.

PR #80, opened 2022-03-24
Small set quotients yield small images of maps to locally small sets

tomdjong:
We can discuss this PR tomorrow.


martinescardo:
Nice!

PR #81, opened 2022-03-27
Small suprema of ordinals

tomdjong:
This is the construction with set quotients. I will formalize the image one later.


tomdjong:
The alternative construction is formalized too now.

PR #88, opened 2022-06-14
Domain theory

tomdjong:
All of the formalization of domain theory, organized into directories and commented.

Because there are now directories, a library file ensures that `agda` can understand the structure when called from the terminal. 
This library file is `typetopology.agda-lib` in the root of the repository.


tomdjong:
@martinescardo When we merge this, we should probably **squash** (all the commits) and merge.

martinescardo:
Wonderful organization and new results. Thanks!

PR #92, opened 2022-07-25
Cat algebra

ettorealdrovandi:
Hello,

Keri (https://github.com/keridangelo) and I have been wanting to implement some Categorical Algebra, in particular Crossed Modules of groups. To that end, but also to have a bit of more theory, we have been putting in facts about group theory: kernels, images, normality, actions, and G-torsors.

Cheers,
--Ettore


martinescardo:
Thank you! I will review this in the first opportunity.

martinescardo:
There are a few things about style that we do in a different way in TypeTopology. However, these things are not documents, and also I think we should probably give liberty to authors to use their own styles. But in any case you are following most of our stylistic choices instinctively. 

There are some things that probably should be changed, in particular copying fragments from other files, but I think we can discuss this later. 

For the moment, I will accept the pull request as is, postponing such discussions for later.

Thanks again for your contribution.

ettorealdrovandi:
We are trying to conform to the style, but of course we didn’t have a set of rules to follow explicitly. I’ll be very happy to discuss the changes. And yes, I realize copying code fragments is bad.  We found ourselves in a bit of a bind: we wanted to have a traditional approach to subgroups, and that exists in SIP-Examples, whereas we had made the (probably unwise) decision to use the interface that is now in Groups.Type.

On Jul 26, 2022, at 10:08, Martin Escardo ***@***.******@***.***>> wrote:



There are a few things about style that we do in a different way in TypeTopology. However, these things are not documents, and also I think we should probably give liberty to authors to use their own styles. But in any case you are following most of our stylistic choices instinctively.

There are some things that probably should be changed, in particular copying fragments from other files, but I think we can discuss this later.

For the moment, I will accept the pull request as is, postponing such discussions for later.

Thanks again for your contribution.

PR #95, opened 2022-09-06
Complete the proof that Patch is Stone

ayberkt:
I've finally completed the proof that Patch is Stone, making a couple of simplifications when porting my Agda code from [`formal-topology-in-UF`](https://github.com/ayberkt/formal-topology-in-UF/).

Let me also note that, currently, the notion of zero-dimensionality doesn't require the basic covering families to be directed. I think this requirement has to be added so that it's compatible with the notion of spectrality. I plan to do this in a future PR.


martinescardo (approved):
Excellent!

PR #97, opened 2022-09-14
Groupoids

ettorealdrovandi:
Add UF.Groupoids defining is-groupoid by analogy with is-set.

We also show that is-groupoid X is equivalent to  X is-of-hlevel 2 and a similar result for is-set.


martinescardo (approved):
Thanks!

PR #99, opened 2022-09-16
Some lemmas needed to prove the universal property of Patch

ayberkt:
Let `L` and `M` be two frames and let `h : | L | → | M |` be a function. Function `h` is said to satisfy the **continuity condition** if *for every `x : L`, compact `b : M` with `b ≤ h(x)`, there is some compact `a : L` such that `a ≤ x` and `b ≤ h(a)`*.

This PR adds the proofs of the following lemmas:

1. Given frames `L` and `M`, with `M` spectral, any monotone function `h : ∣ L ∣ → ∣ M ∣` satisfying the continuity condition is Scott-continuous. (Called `characterisation-of-continuity`.)
1. We now prove the converse: given frames `L` and `M`, with `L` spectral, any Scott-continuous function `h : ∣ L ∣ → ∣ M ∣` satisfies the continuity condition. (Called `characterisation-of-continuity-op`.)
1. As a corollary of (2): given any spectral frame `F`, for every `x, y, : F` and compact `a : F` with `a ≤ x ∨ y`, there exist compact `b, c : F` with `a ≤ b ∨ c` such that `b ≤ x` and `c ≤ y`. (Called `compact-join-lemma`.)

(3) will be used in a crucial way in the proof of the universal property.


ayberkt:
Please ignore this PR. It was branched off branch `patch-zero-dimensional` which was squash-merged into `master` (in #95). I created a new branch, this time branched off `master` after the merge of #95, and cherry-picked the relevant commits from this branch. The PR for that branch (namely #100) should be merged instead of this one.

PR #101, opened 2022-09-27
Reflective Subuniverses

jonsterling:
I have started filling up the "blackboard" with some basics of subuniverses, reflective subuniverses, etc. Not much here yet, but I'll be working on this recreationally over the next few weeks probably. No need to merge right away, but I thought I'd open a draft pull request now in case there is any feedback and suggestions, etc.

### DIFF
--- a/source/Modal/Homotopy.lagda
+++ b/source/Modal/Homotopy.lagda
@@ -42,7 +42,7 @@ homotopy-precomp-by-embedding-is-equiv
  → (precomp-i-is-emb : is-embedding λ (- : X → Y) → - ∘ i)
  → is-equiv (homotopy-precomp f g i)
 homotopy-precomp-by-embedding-is-equiv fe0 fe1 fe2 fe3 f g i precomp-i-is-emb =
- transport is-equiv composite-is-pre-whisker (eqtofun- composite)
+ transport is-equiv composite-is-precomp (eqtofun- composite)
 
  where
   composite : f ∼ g ≃ (f ∘ i ∼ g ∘ i)
@@ -51,8 +51,8 @@ homotopy-precomp-by-embedding-is-equiv fe0 fe1 fe2 fe3 f g i precomp-i-is-emb =
     ● (ap (_∘ i) , embedding-embedding' _ precomp-i-is-emb _ _)
     ● ≃-funext fe1 (f ∘ i) (g ∘ i)
 
-  composite-is-pre-whisker : eqtofun composite ＝ homotopy-precomp f g i
-  composite-is-pre-whisker =
+  composite-is-precomp : eqtofun composite ＝ homotopy-precomp f g i
+  composite-is-precomp =
    dfunext fe2 λ h →
    eqtofun composite h ＝⟨ ap happly (aux h) ⟩
    happly (dfunext fe1 (h ∘ i)) ＝⟨ happly-funext fe1 _ _ (h ∘ i) ⟩
--- a/source/Modal/Homotopy.lagda
+++ b/source/Modal/Homotopy.lagda
@@ -1,0 +1,70 @@
+Jon Sterling, started 5th Oct 2022
+
+This file contains some lemmas about precomposing embeddings onto homotopies.
+
+\begin{code}
+
+{-# OPTIONS --without-K --exact-split --safe --auto-inline #-}
+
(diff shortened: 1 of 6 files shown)


martinescardo:
Nice! It would be good to eventually add a few comments.

martinescardo:
For the sake of consistency with the existing code, would you mind using 1-space indentation for `where`, plus one space within `where`? (Some very old code doesn't conform to that.) Thanks.

source/Modal/Subuniverse.lagda
```diff
@@ -71,8 +71,8 @@ subuniverse-is-reflective P =
 subuniverse-is-replete
  : subuniverse 𝓤 𝓥
  → 𝓤 ⁺ ⊔ 𝓥  ̇
-subuniverse-is-replete {𝓤 = 𝓤} P =
- (A B : 𝓤 ̇)
+subuniverse-is-replete P =
+ (A B : _)
```
plt-amy, line 75: `∀ A B → ...` works too fwiw

jonsterling: I ban that symbol from my code, because it is mathematically weird (based on old school propositions-as-types 🧠 damage IMO)

source/Modal/ReflectiveSubuniverse.lagda
```diff
@@ -0,0 +1,389 @@
...
+    aux =
+     inv-f-sec (f x) ⁻¹ ∙ ap f (inv-f-ret x) ＝⟨ ap (inv-f-sec (f x) ⁻¹ ∙_) (inv-f-sec-coh x) ⟩
+     inv-f-sec (f x) ⁻¹ ∙ inv-f-sec (f x) ＝⟨ trans-sym (inv-f-sec (f x)) ⟩
+     refl ∎
+
+
+  Σ-map-bwd-is-retraction : Σ-map-bwd ∘ Σ-map-fwd ∼ id
+  Σ-map-bwd-is-retraction u = to-Σ-＝ (Σ-map-bwd-is-retraction-pr₁ u , Σ-map-bwd-is-retraction-pr₂ u)
+
+  Σ-map-fwd-is-equiv : is-equiv Σ-map-fwd
```
jonsterling, line 246: Later on I'll move this to the right place; but I was curious, did anyone try proving a lemma like this before (without univalence)? It seemed like it was a lot harder than it should have been, but I wasn't sure if I took a wrong turn somewhere. I was kind of surprised that I needed to use the `equivs-are-haes` fact.

martinescardo: I think what you are looking for is here: https://www.cs.bham.ac.uk/~mhe/TypeTopology/Utilities.PairFun.html
Perhaps I should move this to `UF/`?
This is actually used a lot, particularly in the code for searchable ordinals.

jonsterling: @martinescardo Oh, thanks for the tip! That's just what I needed.

jonsterling: It is a very useful function, and putting it in `UF` does sound like a good idea from my perspective :)

jonsterling (commented):
- [x] Need to make an `index` file

martinescardo (commented):
Jon, please let me know when your code is ready to be reviewed.

jonsterling:
Thanks @martinescardo! I think I'm ready, have at it!

source/Modal/Homotopy.lagda
```diff
@@ -0,0 +1,70 @@
...
+
+homotopy-precomp
+  : {𝓤 𝓥 𝓦 : Universe}
+  → {U : 𝓤 ̇} {X : 𝓥 ̇} {Y : 𝓦 ̇}
+  → (f g : X → Y)
+  → (i : U → X)
+  → f ∼ g
+  → f ∘ i ∼ g ∘ i
+homotopy-precomp f g i h =
+ h ∘ i
```
martinescardo, line 25: Did you really mean to split the line here?

martinescardo (approved):
Nice! Thanks for your contribution. Have you added yourself to the authors index at the top-level README.md file? You should do this in a subsequent pull request. And have you added an import of your new index file to `source/index.lagda`?

PR #104, opened 2022-10-06
Import my module at the top level, update contributors

jonsterling:


source/index.lagda
```diff
@@ -111,6 +111,8 @@ import UF.index
 import Uncatalogued.index
 import Utilities.index
 
+import Modal.index                    -- by Jon Sterling
```
martinescardo, line 114: I'd rather have the imports in alphabetical order. Thanks.

martinescardo (changes requested):

source/index.lagda
jonsterling: oops! Done 😄

martinescardo: Thanks.

PR #105, opened 2022-10-06
Improvements to urls in Modal/Index

tomdjong:
- Improve readability of the doi url of "Modalities in homotopy type theory"
- Add arXiv link to Egbert's thesis.


jonsterling:
thanks!!

PR #106, opened 2022-10-11
Add a generalisation of the proof from the `CantorSearch` module

ayberkt:
This PR adds a generalisation of the proof of the _uniform searchability_ of the Cantor space, showing that the type `ℕ → X` is compact w.r.t. any uniformly continuous predicate `p : (ℕ → X) → 𝟚` and for any compact type `X`. The theorem in the `CantorSearch` module is the special case of this proof where `X := 𝟚`.

### DIFF
--- a/source/TypeTopology/UniformSearch.lagda
+++ b/source/TypeTopology/UniformSearch.lagda
@@ -1,0 +1,10 @@
+Ayberk Tosun
+11 October 2022
+
+In the `CantorSearch` module, the type `ℕ → 𝟚` is proved to be searchable for
+uniformly continuous predicates. In this module, we generalise this to the type
+`ℕ → X` where `X` is an arbitrary compact type.
+
+\begin{code}
+
+\end{code}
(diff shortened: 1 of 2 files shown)


martinescardo (approved):
I wonder if it wouldn't be just as easy to prove a more general fact, using a family `X : ℕ → 𝓤` of searchable types, to show that the product `((n : ℕ) → X n)` is searchable wrt to locally constant functions? By the way, `lc` as an abbreviation is not very good, because we already use `lc` for left cancellability.

ayberkt:
> using a family `X : ℕ → 𝓤` of searchable types, to show that the product `(n : ℕ) → X n` is searchable wrt to locally constant functions?

Right, that shouldn't be hard I suppose. I'll think about that.

> By the way, `lc` as an abbreviation is not very good, because we already use `lc` for left cancellability.

Ah yes, true. Should I expand it out to locally constant or should I use a different abbreviation?

martinescardo:
An abbreviation should be fine.

PR #108, opened 2022-11-22
Ordinals and the cumulative hierarchy

tomdjong:
In constructive set theory an ordinal is usually defined as a transitive set of transitive sets. We consider the subtype 𝕍ᵒʳᵈ of the cumulative hierarchy 𝕍 of set theoretic ordinals in 𝕍 (see Section 10.5 of the HoTT Book). We show that (𝕍ᵒʳᵈ,∈) is a ordinal, in the type theoretic sense, i.e. it is a well-founded, extensional and transitive order. Moreover, we prove that (𝕍ᵒʳᵈ,∈) and the ordinal Ord of type theoretic ordinals are isomorphic.

There are a few minor additions to the files on the general theory of ordinals, including:
- keeping track of the behaviour of transfinite induction;
- two lemmas on describing initial segments of the supremum of ordinals.

Most of the PR is concenctrated in `UF/CumulativeHierarchy.lagda` which introduces 𝕍 and `Ordinals/CumulativeHierarchy.lagda` which contains the result announced at the top.


source/UF/CumulativeHierarchy.lagda
```diff
@@ -0,0 +1,416 @@
...
+  ∙ 𝕍 is set-truncated: for every x, y : 𝕍 and p, q : x ＝ y, we have p ＝ q.
+
+\begin{code}
+
+module _ (𝓤 : Universe) where
+
+ record cumulative-hierarchy-exists : 𝓤ω where
+  field
+   𝕍 : 𝓤 ⁺ ̇
+   𝕍-is-set : is-set 𝕍
```
martinescardo, line 102: I think it is worth adding a remark here, for set theorists. Of course, 𝕍 is a proper class not a set. But also, of course, line 102 is saying that it is a (large) set in the sense of HoTT/UF. I wonder whether we should define `is-large-set`, and use that instead. Or else you should add a comment right at line 102. I probably prefer the first solution.

martinescardo (changes requested):

martinescardo:
You should also update the list of contributors at the main README.md file. Thanks.

source/UF/CumulativeHierarchy.lagda
tomdjong: Nice suggestion, thanks! (I went with the first solution.)

martinescardo:
Thanks for the birthday present!

PR #111, opened 2022-12-16
Exploring Guillaume Munch-Maccagnoni's duploids

jonsterling:
This is based on the work of Guillaume Munch-Maccagnoni.

## Basic concepts

A "deductive system" is a category-like structure removes the associativity law. In a deductive system, we define the following things":

1. A morphism `f` is "linear" when `f; (g; h) = (f; g) h` for all `g,h`.
2. A morphism `f` is "thunkable" when `h; (g; f) = (h; g); f` for all `g,h`.
3. An object is "positive" when every morphism out of it is linear.
4. An object is "negative" when every morphism into it is thunkable.
5. An object is "polarized" if it is either positive or negative (but it could be both).

A "preduploid" is a deductive system in which every object is polarized. A "duploid" is a preduploid in which there are "shifts" between the polarities:

1. An upshift for an object `A` is a negative object `A-` and an invertible linear morphism `force : A- -> A`.
2. A downshift for an object `A` is a positive object `A+` and an invertible thunkable morphism `wrap : A -> A+`.

A "duploid" is a preduploid in which every object has both an upshift and a downshift.

## Relation to adjunctions

From a duploid, it is possible to form **categories** `Pos, Neg` of positive objects and negative objects respectively (in these cases the associativities will work out). Here we have `Down : Neg -> Pos` and this is **left adjoint** to `Up : Pos -> Neg`.

However, if we further restrict these to `PosTh` (the category of positive objects and thunkable maps) and `NegLin` (the category of negative objects and linear maps), the `Down/Up` functors restrict to an adjunction in the opposite direction, in which `Down` is right adjoint to `Up` rather than left adjoint. Let's refer to this as `Up' -| Down'` to avoid confusion with `Down -| Up`.

What is happening here is the following: `Pos` is the Kleisli category for the monad induced by `Up' -| Down'` and `Neg` i the Kleisli category for the comonad of `Up' -| Down'`. Then `Down -| Up` arises as the induced adjunction between the Kleisli categories.

The `Up' -| Down'` adjunction has a special property, namely an "equalizing requirement" that is similar to Moggi's condition on monads in the monadic metalanguage that the units shall be monomorphisms. Duploids are then the same thing as adjunctions satisfying this equalizing requirement. The style of duploids is to derive behavioral/extensional notions of "value" and "stack" (the notions of pure program for positive and negative types respectively) from the behavior of effectful programs, rather than the other way around (as in e.g. Levy's call-by-push-value).

## Goals

One of my goals in formalizing these things from a univalent point of view is to rigorously understand what aspects of the duploid concept are propositional and what parts are structure. This is pretty subtle, as (e.g.) arbitrary isomorphisms in a deductive system are not as well behaved as isomorphisms in a category; for instance, without some associativities, inverses are not uniquely determined.

Furthermore, to be able to state the duploid axiom (closure under upshifts) as a property, we would need to have a **univalent** or saturated notion of preduploid (the current formulation is analogous to a precategory / non-univalent category). I believe that by thinking through the connection to adjunctions, a suitable univalence axiom can be imposed for duploids, but it is not completely clear to me at this time what it will look like. So my goals are to explore these aspects, and develop some of the theory of duploids and denotational semantic of sequent calculus long the way.

I may need to develop some basic univalent category theory in the process, for instance, to have the relationship to adjunctions, etc.

### DIFF
--- a/source/Categories/Adjunction.lagda
+++ b/source/Categories/Adjunction.lagda
@@ -4,49 +4,30 @@ Jon Sterling, started 18th Dec 2022
 
 {-# OPTIONS --without-K --exact-split --safe --auto-inline #-}
 
-open import MLTT.Spartan
 open import UF.FunExt
 
--- TODO: these funext assumptions are getting out of hand;
--- perhaps we should reconsider how these are done?
-module Categories.Adjunction
- (fe1 : funext 𝓥 𝓥)
- (fe3 : funext 𝓥' 𝓥')
- (fe4 : funext 𝓤 (𝓤 ⊔ 𝓥 ⊔ 𝓥'))
- (fe5 : funext 𝓥 𝓥')
- (fe6 : funext 𝓤' (𝓥 ⊔ 𝓤' ⊔ 𝓥'))
- (fe7 : funext 𝓥' 𝓥)
- where
+module Categories.Adjunction (fe : FunExt) where
 
+open import MLTT.Spartan
 open import UF.Base
 open import UF.Equiv
-open import UF.Lower-FunExt
 open import UF.Retracts
 open import UF.Subsingletons
 open import UF.Subsingletons-FunExt
 open import UF.Equiv-FunExt
 
-open import Categories.Category
-open import Categories.Functor
-open import Categories.NaturalTransformation
-
-
-private
- fe0 : funext 𝓤 (𝓤 ⊔ 𝓥)
- fe0 = lower-funext 𝓤 𝓥' fe4
-
- fe2 : funext 𝓤' (𝓤' ⊔ 𝓥')
- fe2 = lower-funext 𝓤' 𝓥 fe6
-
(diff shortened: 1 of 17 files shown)


jonsterling:
Before this gets too enormous to swallow at once, I'm going to add some prose and then mark it ready for review.

martinescardo (approved):
Very nice. Thanks. 

Regarding your question about funext, I would just assume global funext in this case.

I used non-global funext in two situations (1) at the early days of this development, (2) still today when exploring foundational issues (for example univalence implies funext, and other things like that). Because I see what you are doing as "mathematics" rather than "foundations", I would suggest to use global funext (either FunExt or Fun-Ext). But it is up to you. For the moment I am accepting this pull request as it is.

martinescardo:
The html rendering is not updated.

PR #112, opened 2022-12-19
[Categories] Better type inference; add associators and unitors

jonsterling:
When Σ-types are used, Agda is for some reason unable to infer a lot of implicit parameters; this becomes too unwieldy when doing category theory (e.g. natural transformations). I do not fully understand what it is about Σ-types that causes this information-loss, but switching to records solves the problem. In each case that I introduced a record, I proved it equivalent to a Σ-type. This is necessary when characterizing path spaces.

In the course of making this change, it was necessary to define the associators and unitors for functor composition. I think this is because when records are used instead of sums, various type dependencies end up involving the entire data of a functor, rather than just its action on objects and morphisms. (I still don't fully understand why that is, btw — but it seems to *also* be the reason why inference works better when records are used.)

### DIFF
--- a/source/Categories/Adjunction.lagda
+++ b/source/Categories/Adjunction.lagda
@@ -22,12 +22,13 @@ open import Categories.NaturalTransformation fe
 
 module adjunction-of-precategories (𝓒 : precategory 𝓤 𝓥) (𝓓 : precategory 𝓤' 𝓥') where
  open functor-of-precategories
+ open natural-transformation
 
  private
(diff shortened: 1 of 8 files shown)


martinescardo (approved):
It's very annoying that records can't be treated uniformly like Sigma types. I think that there is a tactic in the cubical library to produce Sigma types from record types and automatically show that they are definitionally equivalent. Perhaps we could adopt this in TypeTopology.

PR #118, opened 2023-02-17
Dyadic Rationals

adsneap:
This pull request contains a formalisation of dyadic rationals. This includes addition along with associativity proofs, multiplication along with associativity and commutativity proofs and negation.

It also includes a new exponents file for integers, which is used in the dyadic rationals files.

PR #119, opened 2023-02-23
Fulfil requests

adsneap:
This branch implements the requested changes to the previous pull request: https://github.com/martinescardo/TypeTopology/pull/118

The two files which implement exponentiation are both renamed as suggested. 

I have made many formatting changes, ensuring that every line is 80 characters or less.


martinescardo (approved):
Looks much better now - thanks!

PR #120, opened 2023-02-24
[Draft] working on dcpo presentations

jonsterling:
This is a draft PR containing @Trebor-Huang's work on exploring dcpo presentations, roughly following [Jung, Moshier, and Vickers](https://www.sciencedirect.com/science/article/pii/S1571066108004076). We will mark it as such when it is ready for review, but preliminary feedback is welcome along the way at any time.


source/DomainTheory/Presentation/Type.lagda
```diff
@@ -0,0 +1,77 @@
...
+open import DomainTheory.Basics.Dcpo pt fe 𝓥
+open import DomainTheory.Basics.Miscelanea pt fe 𝓥
+
+module _
+  (G : 𝓤 ̇)  -- Generators
+  (_≲_ : G → G → 𝓣 ̇)
+  where
+
+  Cover-set : 𝓤 ⊔ 𝓥 ⁺ ⊔ 𝓦 ⁺ ̇ -- This one has spurious assumptions
+  Cover-set = G → {I : 𝓥 ̇} → (I → G) → Ω 𝓦
```
jonsterling, line 34: I am wondering whether this is a workable way to formulate the set of covers in a univalent setting. In particular, I think we would need it to be a family of types rather than a family of propositions. But maybe there is some better formulation anyway.

Trebor-Huang: There are two powersets in the original definition. I changed the inner one but not the outer one, because I was not so sure about how the outer one is used.

martinescardo:
Any progress with this?

Trebor-Huang:
I'm about to come back to this as summer holiday is coming.

martinescardo:
Is there an intention to continue this work?

ayberkt:
@jonsterling @Trebor-Huang Do you plan to finish this?

martinescardo:
Is there any intention to continue with this?

jonsterling:
I don't have plans to continue with this.

PR #124, opened 2023-03-25
Cantor's theorem for embeddings; cardinals

jonsterling:
I was curious about another other statement of Cantor's theorem, for embeddings `P(A) -> A`. This follows, impredicatively, from the retract version --- I wasn't sure if there was, however, an easier way to deduce it. I use this result to show that for any cardinal, there is another cardinal strictly larger than it. I am going through all sorts of painful contortions involving resizing, so the code could definitely use some cleanup and improvement; suggestions welcome.


source/Various/LawvereFPT.lagda
```diff
@@ -191,6 +197,67 @@ As a simple application, it follows that negation doesn't have fixed points:
...
+Mathematics, via the nLab: https://ncatlab.org/nlab/show/Cantor%27s+theorem.
+
+\begin{code}
+
+ cantor-theorem-for-embeddings
+  : FunExt
+  → PropExt
+  → Propositional-Resizing
+  → (A : 𝓤 ̇ )
+  → (ϕ : (A → Ω₀) → A)
```
jonsterling, line 213: I'd like to state this with `\Omega U` instead; this makes the inside of the proof a bit more complex, but not impossibly so.

source/Cardinals/Successor.lagda
```diff
@@ -0,0 +1,149 @@
...
+open import Cardinals.Type st
+open import Cardinals.Preorder fe pe st pt
+
+import UF.Logic
+
+open set-truncations-exist st
+open propositional-truncations-exist pt
+open UF.Logic.AllCombinators pt (fe _ _)
+
+resize-Ω : Ω 𝓤 → Ω 𝓥
```
jonsterling, line 42: all the code involving `resize-Ω` should be moved to the `UF.Size` module

PR #125, opened 2023-04-08
Code review of Rationals (and pre-requisite files of rationals), removal of function extensionality dependence from rationals

adsneap:
This pull request modifies a large number of files, with the following changes:

- Remove function extensionality from the rationals. This is accomplished by introducing a new "coprime" relation on natural numbers, testing if the hcf of two numbers is equal to one. Since the naturals are a set, this relation does not require function extensionality. By proving that the previous definition of coprime implies the new one, only minor modifications are required in proofs which use the old definition of "coprime".
- Reviews the code of the Integers files, improving some proofs where possible, and adhering to the 80 character limit.
- Reviews the code of the Rationals files, improving many proofs, making them clearer (but not always shorter in terms of lines), and adhering to the 80 character column limit except in specific circumstances.
- Removes the scratchpad "ContinuousExtensionTheorem" file. The ideas in this file were implemented in "MetricSpaces/Extension.lagda", and so this file is no longer useful.


source/Rationals/Addition.lagda
```diff
@@ -10,184 +10,187 @@ properties of addition.
...
-1/2+1/4 : Fun-Ext → 1/2 + 1/4 ＝ 3/4
-1/2+1/4 fe = equiv→equality fe ((pos 1 , 1) ℚₙ+ (pos 1 , 3)) (pos 3 , 3) refl
+1/2+1/4 : 1/2 + 1/4 ＝ 3/4
+1/2+1/4 = equiv→equality ((pos 1 , 1) 𝔽+ (pos 1 , 3)) (pos 3 , 3) refl
 
 \end{code}
 
-For the following code, the flag --experimentay-lossy-unification must
-be added, otherwise the file compiles infinitely.
+For the following code, the flag lossy-unification must be added, otherwise the
```
martinescardo, line 136: Change to "otherwise the type checking takes too long to complete".

martinescardo (approved):
I am approving, but I am suggesting a few minor changes that you can make later. Thanks a lot for doing this.

adsneap:
> I am approving, but I am suggesting a few minor changes that you can make later. Thanks a lot for doing this.


In regards to the capitalisation of Bezout, I agree. Does this also apply to instances of proof names? E.g "Euclid" and "Bezout" are both lowercase in the current version Agda code.

martinescardo:
As far as the names of functions is concerned, I am happy for you to use lower case for them. But in comments they should appear in upper case, like Andrew Sneap as as opposed to andrew sneap. :-)

PR #126, opened 2023-04-20
Continuous Extension Theorem

adsneap:
This pull requests adds a new Extension file to the Dedekind Reals folder, which defines an extension function on rational functions, lifting them to real functions. It is proved that this extension function is actually an extension, by proving that for rational inputs the extension agrees with the rational function.

The changes requested in the previous pull request are included.

The new import is included in the index, and necessary lemmas are added to relevant pre-requisite files.


martinescardo (approved):

martinescardo:
Thanks. This is great and really nice.

PR #127, opened 2023-04-24
Fix typo in extension file, add reals abs function

adsneap:
This pull request fixes the typo at the top of the extension file, and adds one more example (absolute value) of a uniformly continuous extension function extended from rationals to reals.


martinescardo:
Thanks!

PR #131, opened 2023-05-19
adapted WithoutOracle so that it uses SystemT and Internal so that it uses WithoutOracle

vrahli:
adapted WithoutOracle so that it uses SystemT.
adapted Internal so that it uses WithoutOracle.


martinescardo (approved):
Thanks!

PR #134, opened 2023-06-02
Refactor the `InitialFrame` module as to use propositional extensionality instead of full univalence (resolves #69)

### LINKED ISSUE #69: Refactor `InitialFrame` to avoid using full univalence
Propositional + functional extensionality should be sufficient as @martinescardo pointed out.

ayberkt:
Resolves #69 by implementing the change discussed there.


martinescardo:
The check failed.

PR #136, opened 2023-06-05
Typos and infimum proof

Seiryn21:
I did the correction you suggested, and added the proof ε is the infimum. But i'm not sure where the proof should go, as it feel a little stranded at the end of the file


martinescardo:
Oh, there is one thing. The arrows in multiple-line definitions should go at the beginning, not at the end, in the `TypeTopology` convention. Would you mind fixing this? Thanks.

martinescardo:
One more, minor thing. Would you mind adding full stops at the end of your comments? Thanks!

PR #138, opened 2023-06-07
Add additional domain-theoretic formalization targets

tomdjong:


ayberkt:
@tomdjong Another thing that I would like to see in the domain theory formalisation is the definition of the Scott topology of a domain and the proof that the Scott-opens form a frame. I intend to do this myself at some point so you can add it to your list of domain theory to-dos if you want. One can even show that the Scott topology of a continuous dcpo is spectral with the definitions in the locale theory library we currently have.

tomdjong:
@ayberkt I agree this would be very nice! I'll make a PR adding these items with a note that if someone wishes to work on them, then they should reach out to you to discuss. OK?

ayberkt:
@tomdjong 

> I'll make a PR adding these items with a note that if someone wishes to work on them, then they should reach out to you to discuss. OK?

Sounds great, thanks!

PR #145, opened 2023-06-11
Internal correctness

vrahli:
still working on the lambda case


martinescardo:
Probably some code there about order and Fin should be moved elsewhere in the future.

(Also, I don't like too much using mathematical types as name of variables. Some people like it, but I think it is very confusing. We should discuss this. For now it's fine.)

PR #151, opened 2023-06-15
Internal correctness

vrahli:
Made some progress on the lambda case.
Moved files around as discussed with Martin, Ayberk, and Bruno.


martinescardo:
Great! Maybe create `index.lagda` in each of the subdirectories, and import these two from the `index.lagda` at `EffectfulForcing`

PR #153, opened 2023-06-20
Cleanup

jonsterling:
- fixed some formatting in `Dominance.Lifting`
- added my `Coslice` module to the root index


martinescardo (approved):
Thanks - will merge when the checks pass.

PR #154, opened 2023-06-23
Internal correctness

vrahli:
A big chunk of the proof should be there now, but we now have to clean up the code.


martinescardo (approved):
Nice! Vincent explained the main ideas in his office to me a few hours ago.

PR #158, opened 2023-06-29
Moved unnecessary code to a separte file and did a bit of cleanup

vrahli:


vrahli:
We've removed a file we're now working on in a different branch.

PR #167, opened 2023-07-28
The type of iterative sets is a retract of the type of multisets

tomdjong:
Also updated the comments on Ord being a retract of the type of iterative sets.


tomdjong:
@martinescardo You broke something in https://github.com/martinescardo/TypeTopology/commit/034d9635b82118f92b396fb4a4d2e89e460c5e31. If you fix it, I can make the PR pass the check as well by pulling `master`.

martinescardo:
It should be fixed now.

PR #170, opened 2023-08-07
Prove the correctness of the internal modulus operator

ayberkt:
The operator in question is called `internal-mod-cont`. Its specification is given by `internal-mod-cont-correct`.


martinescardo (approved):
Great!

martinescardo:
Minor comment. There are sometimes two blank lines before `\end{code}`. I think the unspoken convention is only one blank line. Also after `\begin{code}`.

ayberkt:
> There are sometimes two blank lines before `\end{code}`.

Oh thanks for pointing this out @martinescardo. This is just me not being careful.

Issue #168, opened 2023-08-02
Decouple the small basis condition from the definition of spectrality


ayberkt:
A long-standing TODO, discussed with @martinescardo before.

ayberkt:
Fixed in #180.

PR #180, opened 2023-09-08
Fix Issue #168 and do a lot of refactoring

ayberkt:


ayberkt:
The idea of this PR was to fix Issue #168. However, it ended up containing a lot more than that.

1. The `CompactRegular` module has been broken down into a lot of smaller modules (around 20).
1. The definition of spectrality has been updated and some things about that have been clarified (in module `SmallBasis`). We now have the equivalences between various ways of defining spectrality.
1. I added a new proof that the type of spectrality structures on a large and locally small locale has _split support_.

There are some more things implemented in this PR, and it would be hard to summarise everything here.

ayberkt:
It might be a good idea to look at this together @martinescardo .

martinescardo (approved):
Thanks a lot for this refactoring & mathematical improvement.

I understand this because we discussed it before you implemented it. But other readers would benefit from lots of more comments. The purpose of TypeTopology is not to "certify" truth, but to record and explain mathematical knowledge. This is particularly important for things that are not already in the literature because they are new. So please do add plenty of comments in the future. They are just as useful, if not more useful, than the Agda code.

ayberkt:
@martinescardo 

> The purpose of TypeTopology is not to "certify" truth, but to record and explain mathematical knowledge.

I completely agree. The development went pretty fast so far, and lots of breaking changes happened before I had the chance to properly clean things up which is why there are lots of uncommented modules in the development.

I created issue #190 to track the progress on this. It contains the list of modules whose literate parts I will review and improve (gradually as I work on other things).

The idea is that, by the time I’m done with this development, each module will be a literate program that can be read cover to cover. Breaking the `CompactRegular` module into lots of smaller modules was the first step of this cleaning.

PR #187, opened 2023-09-17
Add `Makefile` generated from the dependency graph of `index.lagda`

ayberkt:


martinescardo (approved):
Wonderful!

Could you add one or two lines of instructions to the make file next time? (Such as how to exploit parallelism with `-j`.)

PR #194, opened 2023-09-25
Fin n is an ordinal

Seiryn21:


source/Ordinals/Fin.lagda
```diff
@@ -0,0 +1,52 @@
+Alice Laroche, 25th September 2023
+
+Fin n is an ordinal
```
martinescardo, line 3: I suggest to write "Fin n is an ordinal under its natural order."

martinescardo (approved):
Thanks!

PR #198, opened 2023-09-29
Update defn. of point and prove the equivalence between CPFs and frame homomorphisms into `Ω`

ayberkt:


ayberkt:
This PR adds the proof discussed at the meeting on 2023-09-27.

It also adds record-based definitions for the notion of point to avoid using projections.

PR #200, opened 2023-10-02
Experiment with redefining path concatenation

tomdjong:
Not all files are fixed, but the current changes should give a pretty clear picture of what would be needed.


tomdjong:
It would be nice if we could run the TypeChecking action while keeping this a draft PR.
@ayberkt Do you happen to know how to do this?

martinescardo:
Thanks for sharing this. I think I would prefer to leave things as they are, given what I see.
However, we may wish, in the future, to try using `opaque` to get better goals, which was the point of this exercise.

jonsterling:
sounds good, and agreed! Thanks for the experiment and I look forward to trying "opaque" once 2.6.4 is out.

PR #201, opened 2023-10-02
Define the locally small Scott locale of an algebraic dcpo

ayberkt:


tomdjong:
Nice to see more topology!

Here are my comments:

#### `Locales/ScottLocale/Definition.lagda`
1. It is probably helpful to be explicit about the universe parameters of the dcpo 𝓓.
1. It would similarly be helpful to include the special case of a single universe parameter at the end of the file:
   ```
   ScottLocale' : (𝓓 : DCPO {𝓥 ⁺} {𝓥})
                → propext 𝓥
                → Locale (𝓥 ⁺) (𝓥 ⁺ ) 𝓥
   ScottLocale' 𝓓 pe = ScottLocale
    where
     open DefnOfScottLocale 𝓓 𝓥 pe
   ```
   This is easier to digest than the more general case (which is more useful of course!), while it still illustrates the fact that the resulting locale is not locally small.

#### `Locales/ScottLocale/ScottLocalesOfAlgebraicDcpos`
1. Your comment says:

   > If one starts with an algebraic dcpo, one can ensure that the resulting Scott locale is small by quantifying over only the basic/compact opens.

   But surely it should be "the resulting Scott locale is *locally* small"?
1. Your comments mention structurally algebraic dcpos, but actually you are using dcpos with a small compact basis. These things are not equivalent (at least as far as I know), so it would be good to fix the comments.
1. The constructions of this file are less general than the one discussed above, as you do use a single universe parameter here. I can see why, but it's maybe worth a comment?
1. This file repeats lots of things established in the other file, like proving distributivity. Would it not be easier to instead show that ⊆ and ⊆ₛ are (pointwise logically) equivalent and transfer the proofs along? This might also be clearer to any readers: you're not really constructing an entirely different locale (the carrier stays the same for example), you're just showing that the order is equivalent to one with small values.

I hope these comments are helpful. I am happy to discuss more, either here or on Discord.

source/Locales/ScottLocale/ScottLocalesOfAlgebraicDcpos.lagda
```diff
@@ -0,0 +1,361 @@
...
+This module contains the definition of the Scott locale of a large and locally
+small _algebraic_ dcpo, using the definition of algebraic dcpo from the
+`DomainTheory` development due to Tom de Jong.
+
+If one starts with an algebraic dcpo, one can ensure that the resulting Scott
+locale is small by quantifying over only the basic/compact opens. This is the
+difference between the construction in this module and the one in
+`ScottLocale.Definition`
+
+TODO: in the future, it would be good to call the other module something
```
martinescardo, line 15: Is this TODO done?

source/Locales/ScottLocale/ScottLocalesOfAlgebraicDcpos.lagda
```diff
@@ -0,0 +1,361 @@
...
+ 𝕒 = structurally-algebraic-if-specified-small-compact-basis 𝓓 hscb
+
+\end{code}
+
+We denote by `I` the index type of the basis and by `β` its enumeration
+function.
+
+\begin{code}
+
+ I = pr₁ hscb
```
martinescardo, line 95: Could we name the projections for clarity and modularity and portability?

martinescardo (changes requested):
Thanks. Tom's comments are useful. Please address them.

tomdjong (changes requested):
As commented.

ayberkt:
Thanks for the comments @tomdjong!

---

  ## `Locales.ScottLocale.Definition`

- Special case for the locally small case.

  > It would similarly be helpful to include the special case of a single
  > universe parameter at the end of the file

  Agreed, and fixed in 66ebe048.
-
  > It is probably helpful to be explicit about the universe parameters of the
  > dcpo 𝓓

  What exactly do you mean by this? Is there a version of `DCPO` that takes the universe parameters as explicit arguments?

## `Locales/ScottLocale/ScottLocalesOfAlgebraicDcpos`

- > But surely it should be "the resulting Scott locale is locally small"?

  Definitely. Fixed now.

- > Your comments mention structurally algebraic dcpos, but actually you are using dcpos with a small compact basis. These things are not equivalent (at least as far as I know), so it would be good to fix the comments.

  I have now fixed the terminology. I had forgotten what “structurally algebraic” means, and revisited the thesis. The source of confusion is that, in my terminology, I use terms like “spectrality structure” to refer to basis structures witnessing spectrality, but the structure in this context refers to the structure of the covering families and not that of the basis itself. I’ll be more careful about this distinction in the future.
  
- > The constructions of this file are less general than the one discussed above, as you do use a single universe parameter here. I can see why, but it's maybe worth a comment?

  That's a very good point. The issue here is that, after writing the general construction in `ScottLocale.Definition`, I realised that it's pretty much useless — because we almost never want to work with lattices that are not locally small. This is why I wrote the new one copying over a lot of things from `ScottLocale.Definition`, thinking perhaps I might get rid of that altogether.

  I haven't yet decided what the relationship between the two modules should be. I will think about the common structure they share, and perhaps refactor the common part into a separate module, of which they both will be instances. It might also make sense to define a general notion of locale of subsets, because that's the more general setting in which facts like `distributivityₛ` should be living. Because this requires a bit of thinking, I have postponed doing this before sending this PR as I'm trying to avoid going down the rabbit hole of endless yak shaving and refactoring. This is why I added a short comment about this in the module, but I didn't bother to explain all this in detail, which was confusing. I have now updated the comment to explain this.

source/Locales/ScottLocale/ScottLocalesOfAlgebraicDcpos.lagda
ayberkt: No, see the discussion in [my response](https://github.com/martinescardo/TypeTopology/pull/201#issuecomment-1746602574) to @tomdjong's comments. I will do this in a future PR.

I have updated the comments to explain this in more detail.

source/Locales/ScottLocale/ScottLocalesOfAlgebraicDcpos.lagda
ayberkt: The reason I used a projection here was actually to _avoid projections_ in the module. I couldn't find a standard projection for this, and I didn't want to modify the `DomainTheory` modules. I'm not sure what the best thing to do here is and I always think about this.

I will look at `DomainTheory.BasesAndContinuity.Bases` more carefully to find out if there's a record-based definition that I can use for this purpose, but that might not be the case. In situations like these, are you fine with me adding record-based definitions for things (cc @tomdjong)?

tomdjong: It's fine with me to add some records.

tomdjong:
> What exactly do you mean by this? Is there a version of `DCPO` that takes the universe parameters as explicit arguments?
> 

No. All I meant was that the *comment* should remind the reader that the dcpo has carrier in U and order taking values in T.

tomdjong:
As for
> I haven't yet decided what the relationship between the two modules should be. I will think about the common structure they share, and perhaps refactor the common part into a separate module, of which they both will be instances.

I think my suggestion (item 4) above would work well.

ayberkt:
@tomdjong 

> I think my suggestion (item 4) above would work well.

I have now implemented your suggestion in 8d35d99b.

I think an even clearer formulation would be to write down a definition of local smallness for a locale that says something like:

> Locale $X$ is locally small if its underlying poset is isomorphic to to some $(\mathcal{U}^+, \mathcal{U})$-poset

and then prove explicitly that the assumption of algebraicity of the dcpo allows one to construct this poset here. But for now, getting rid of the extremely confusing duplication of definitions here is good enough, I think.

> No. All I meant was that the comment should remind the reader that the dcpo has carrier in U and order taking values in T.

 Ah I see. I had a detailed comment on the module parameters but it seems like I didn’t say anything about the dcpo’s universes there. I have now updated that comment as follows:
 

```
 We carry out the construction in the following submodule which is parameterised by

  1. a dcpo `𝓓` whose (a) carrier set lives in universe `𝓤`, (b) whose relation
     lives in universe `𝓣`, and (c) whose directed joins are over families with
     index types living in universe `𝓥`.
  2. a universe `𝓦` where the Scott-open subsets are to live,
  3. an assumption that `𝓦` satisfies propositional extensionality.
```

tomdjong:
RE: local smallness; you might want to compare with my definition over at `DomainTheory.Basics.Miscelanea`. If you have a poset with 𝓥-small suprema, then you ask the order to take values in 𝓥 up to equivalence.
But I agree that avoiding duplication is more important and enough for now.

RE: the comment; this is indeed what I had in mind :+1:

tomdjong:
@ayberkt Are things ready for another review now?

ayberkt:
> @ayberkt Are things ready for another review now?

@tomdjong Let me do a bit more yak shaving before we merge this. I’ll let you know.

martinescardo:
What is the status of this?

ayberkt:
> What is the status of this?

@martinescardo It’s almost ready to merge, though I want to look at it one more time because I made quite a few changes rather fast. I will do this as soon as I’m done with the talk that I’m giving today at 4pm.

tomdjong:
@ayberkt Thanks! I'll do a review on Tuesday after our meeting, OK?

source/Locales/ScottLocale/ScottLocalesOfAlgebraicDcpos.lagda
ayberkt: >  It's fine with me to add some records.

I will do this in a future PR @martinescardo. I don’t want to add too many tangential changes into this PR. I 100% agree that I should completely avoid projections. If there were a well-principled way for me to avoid projections (i.e. enforcing this using the type system), I would use it.

In the presence of an ML-like module system, for example, I would keep the `Σ`-based definition abstract, so that projections are literally not available to me, and I would expose only the equivalence with the record-based definition, so that this is the only way to make use of an inhabitant of the `Σ`-based definition. In Agda, one could use typeclasses to do this perhaps, but I’m not sure if it’s a good idea.

ayberkt:
> @ayberkt Thanks! I'll do a review on Tuesday after our meeting, OK?

Sounds good.

ayberkt:
@martinescardo You can merge this now.

martinescardo (approved):

tomdjong (approved):

Issue #189, opened 2023-09-19
Add proof that the type of small compact bases for large and locally small dcpos has split support


ayberkt:
Dcpo analogue of a proof implemented in #180. Discussed with @tomdjong.

ayberkt:
My suggestion in this issue was implemented by @tomdjong in PR #202. The proof lives in module `DomainTheory.BasesAndContinuity.CompactBasis`.

PR #202, opened 2023-10-04
A dcpo with an unspecified small compact basis is structurally algebraic

tomdjong:
Result due to @ayberkt.

Once I cleaned up the code, I will merge this myself.


ayberkt:
Great! By the way, if you write “fixes #189” either in a commit message or in the PR description, the merging of this PR will automatically result in issue #189 closing.

Could we also formulate this explicitly using the terminology of “having split support” from `UF.KrausLemma`? I was happy to see that the vocabulary to express this situation had already been formulated by [Kraus, Escardó, Coquand, and Altenkirch](https://doi.org/10.23638/LMCS-13(1:15)2017).

martinescardo:
The terminology "split support" comes from topos theory, actually. The support of an object X is the image of the unique map `X->1` (and this is the same as the propositional truncation of X). And split is like as usual in category theory.

source/DomainTheory/BasesAndContinuity/CompactBasis-SplitSupport.lagda
```diff
@@ -0,0 +1,357 @@
...
+  has-specified-small-compact-basis 𝓓 ⇔ has-unspecified-small-compact-basis 𝓓
+ specified-unspecified-equivalence =
+    ∣_∣
+  , specified-small-compact-basis-has-split-support
+
+\end{code}
+
+We stress that the above cannot be promoted to an equivalence of types as there
+could be several small compact bases for 𝓓. In a sense, the above says that
+there is a canonical one though (which is given by all compact elements).
```
martinescardo, line 342: Any basis of an algebraic dcpo has to include all compact elements. You probably mean something else.

tomdjong:
This fixes issue #189.

source/DomainTheory/BasesAndContinuity/CompactBasis-SplitSupport.lagda
tomdjong: The definition of small compact basis allows for some redundancy, e.g. by taking two copies of all compact elements. In any case, I'll try to clarify the comment, but I need to think a bit about that. Thanks.

martinescardo: Shouldn't there be a unique basis which is an embedding, even without requiring smallness conditions?

tomdjong: Yes, that's right. In fact, I explain to include this as part of an improved explanation in a commit later today.
Once I'm done, I'll ping you here to get your thoughts on it. Is that all right?

tomdjong:
@martinescardo @ayberkt I had another go at explaining what's going on. It can probably be improved still, so your comments (at your convenience) are very welcome!

tomdjong:
@martinescardo @ayberkt I had another go at explaining what's going on. It can probably be improved still, so your comments (at your convenience) are very welcome!

martinescardo (approved):
Wonderful! I am happy with the comments now.

ayberkt:
@tomdjong This is perfect, thanks again. I have only two comments.

#### Comment 1

I have already [formalised the same proof][0] for spectral locales. I would appreciate if you could add a reference to this.

#### Comment 2

I think it could also be useful to mention the context in which this observation originated.

For some constructions, both in locale theory and domain theory, we want to use the structure of the basis (by which I mean the structure of the family giving the basic opens, not that of the covering families) in a crucial way.

1.  One example is the patch locale, whose local smallness depends crucially on the ability to quantify over all basic opens. This is because for Scott-continuous nuclei on a spectral locale, ordering two nuclei pointwise using compact opens is equivalent to the usual pointwise order (that quantifies over arbitrary opens). Without this, patch would not have been locally small — and without a locally small patch, our proof (with @martinescardo and @IgorArrieta) of the universal property would not have (necessarily) worked.
1. Another example is the Scott locale of an algebraic dcpo whose local smallness uses an order quantifying over basic opens.

In both of these examples, I wanted to quantify over the basic opens, despite only _knowing merely that the structure in consideration exists_. Luckily for me, this is possible, which is the result added in this PR. As a follow-up to PR #201, for example, I will define a version of the locally small Scott locale of an algebraic dcpo that assumes only an unspecified compact basis, and I will use the result here to connect it to the result that assumes a specified basis.

Of course, it’s not a good idea to go into so much detail, but I think the motivation for this observation deserves a mention.

[0]: https://www.cs.bham.ac.uk/~mhe/TypeTopology/Locales.SmallBasis.html#23217

tomdjong:
> **Comment 1**

Thanks for the pointer! I'll add a reference as a comment at the top. I was not aware earlier, sorry.

> **Comment 2**

Agreed. I only point out that for establishing local smallness, and hence for both of the examples you mentioned, full univalence is not needed (since you're resizing a proposition), so the result in the PR is in a sense 'too strong' for these purposes.

PR #206, opened 2023-10-12
Continuous and algebraic dcpos impredicatively

9Y0:
If we assume `propositional-resizing (𝓥 ⁺ ⊔ 𝓣) 𝓥`, we can recover continuous and algebraic structures on `𝓓 : DCPO {𝓥} {𝓣}` from their respective properties. Note that the universes of the carrier types has to be `𝓥` for this to work. @nmvdw suggested me to formalize this in Type Topology, after he formalized it in Unimath (https://github.com/UniMath/UniMath/pull/1762). The Type Topology formalization is slightly different, as we have to be more careful with universes.

I'm fairly new to using Type Topology, so any feedback on incorrect style/naming, would be greatly appreciated.


source/DomainTheory/BasesAndContinuity/Continuity.lagda
```diff
@@ -505,3 +505,239 @@ module _
   ∥∥-functor structural-continuity-of-dcpo-preserved-by-continuous-retract
 
 \end{code}
+
+If we assume propositional resizing, we can recover a continuity/algebraic
```
martinescardo, line 509: Please add your name and date here.

source/DomainTheory/BasesAndContinuity/Continuity.lagda
```diff
@@ -505,3 +505,239 @@ module _
...
+        family-is-inhabited =
+         ∥∥-functor
+          approximating-family-index-to-index
+          (inhabited-if-Directed 𝓓 _ (approximating-family-is-directed x))
+
+        family-is-semidirected : is-Semidirected 𝓓 (family x)
+        family-is-semidirected (y₁ , y₁≪x) (y₂ , y₂≪x) =
+         ∥∥-rec₂ ∃-is-prop f h1 h2
+         where
+           f : Σ i ꞉ index-of-approximating-family x , y₁ ⊑⟨ 𝓓 ⟩ approximating-family x i
```
martinescardo, line 569: Indentation in `TypeTopology` is by 1 space.

martinescardo:
Thanks for your contribution. I'll ask @tomdjong to review it.

martinescardo:
You should also add yourself to the list of contributors in the top `README.md` file in this pull request.

9Y0:
I also figured out that there is another choice possible for `structurally-algebraic-if-algebraic`. In this PR I gave the chosen compact family the index type `Σ y ꞉ ⟨ 𝓓 ⟩ , is-compact 𝓓 y × y ≪⟨ 𝓓 ⟩ x` for `x : ⟨ 𝓓 ⟩`. However, it's also possible to get an algebraic structure if we take `Σ y ꞉ ⟨ 𝓓 ⟩ , is-compact 𝓓 y × y ⊑⟨ 𝓓 ⟩ x` as the index type. The rest of the construction is then almost the same.

Would either one of these constructions be preferred over the other?

tomdjong:
Hi Simcha (@9Y0), I'm just writing a quick message here to say thanks and to let you know that I'll look at this properly soon. But today (and tomorrow) are very busy already, sorry about that.

source/DomainTheory/BasesAndContinuity/Continuity.lagda
```diff
@@ -505,3 +505,241 @@ module _
...
+   ; approximating-family-is-directed = family-is-directed
+   ; approximating-family-is-way-below = family-is-way-below
+   ; approximating-family-∐-＝ = family-∐-＝
+   }
+   where
+    index : ⟨ 𝓓 ⟩ → 𝓥 ̇
+    index x = Σ y ꞉ ⟨ 𝓓 ⟩ , resize psz (y ≪⟨ 𝓓 ⟩ x) (≪-is-prop-valued 𝓓)
+
+    make-index : {x : ⟨ 𝓓 ⟩} (y : ⟨ 𝓓 ⟩) → y ≪⟨ 𝓓 ⟩ x → index x
+    make-index y p = y , (to-resize psz _ (≪-is-prop-valued 𝓓) p)
```
tomdjong, line 537: You can drop the outer parentheses here.

9Y0:
> Hence, it would make sense to refactor your code to assume that the dcpo is locally small.

If I would assume this, do you still think it's better to move the constructions to a new file `ContinuousImpredicative`, or should they stay in `Continuity` in that case?

tomdjong:
Yes, it would still be better to move to a new file, because you assume the carrier to be small as well, which never happens predicatively. Also, it's clearer that it's a separate result from a new contributor if it's a new file.

(In a sense, the files are organised more like little articles than as a formalisation library.)

source/DomainTheory/index.lagda
```diff
@@ -58,18 +58,20 @@ import DomainTheory.Basics.WayBelow        -- (7)
    split support (j.w.w. Ayberk Tosun; added 4 & 5 October 2023)
 3. The theory of continuous/algebraic dcpos
 4. A discussion on possible definitions of continuous dcpo
-5. The Ind-completion is used to discuss the notion of (structurally/pseudo-)
+5. Continuous/algebraic dcpos in an impredicative setting
```
tomdjong, line 61: @9Y0 Could you please add something like "(Added by Simcha van Collem on ...)" to clarify that this is not part of the original development? Thanks!

tomdjong:
Thanks for the improvements!

9Y0:
Thanks for all the feedback! I considered using the `reindexed-family` lemmas, but `reindexed-family-sup` isn't ideal when reasoning about `∐`, and the semidirectness proofs didn't really need much change when reasoning with either the small or the big index type, so I think it's fine like this.

PR #207, opened 2023-10-12
Absolute Difference lemmas and refactoring

lane-core:
I expanded upon the Absolute Difference module I created last September and also refactored the type definitions. 

Also, I have a new module to submit once these changes are committed. I developed a partial order type for Nat that uses an identity type to generate the family of inequalities in Nat. I will submit that as soon as this PR has been reviewed/merged.


lane-core:
I need to redo this PR, some changes are necessary

PR #208, opened 2023-10-13
Define the notion of rank for System T types

ayberkt:
Something that we have discussed with @vrahli and may potentially be useful in the future.


ayberkt:
I would like to request reviews from @brunorochapaiva and @vrahli, but it seems like they do not have the permission.

PR #209, opened 2023-10-13
Absolute Difference lemmas and refactoring

lane-core:
I expanded upon my Absolute Difference module, refactored to explicitly specify used lemmas from other modules, and added one lemma to the Order file.


martinescardo (approved):
Thanks.

PR #172, opened 2023-08-16
Thesis pr

tnttodda:

PR #220, opened 2023-11-26
Remove early thesis draft

tnttodda:
An early draft of my thesis is live on typetopology - it would be best to remove this and update the original thesis-pr with my corrected thesis once this is ready


ayberkt:
@tnttodda Do you mean you’re planning to abort draft PR #172 and start a new one?

PR #225, opened 2024-01-17
Add index for the Patch paper and do a bit of refactoring

ayberkt:


ayberkt:
Closing this as there is no intention to complete it. I will, however, be compiling a comprehensive index for the thesis.

PR #226, opened 2024-01-18
Carriers of pointed dcpos are algebraically injective

tomdjong:
@martinescardo As discussed this morning.


martinescardo (approved):
You may wish to emphasize, as discussed this morning, that this doesn't mean that pointed dcpos are injective in the category of dcpos and continuous maps (and they are not). What I am saying is implied by the fact that you are talking about their carriers, but perhaps it is worth emphasizing this. I leave this to a future pull request.

PR #228, opened 2024-01-25
Mostly complete the removal of function extensionality

ayberkt:
This PR fixes issue #74.


ayberkt:
@brunorochapaiva: I have now completed the removal of function extensionality on which we were working together (on 2023-11-08).

ayberkt:
@martinescardo: Should we merge this or should we first try to remove function extensionality in the case of uniform continuity as well? This PR completely removes function extensionality for computing moduli of continuity but not for moduli of uniform continuity (on the Cantor space).

I think it’s fine to merge this now (clearly marking that it doesn’t remove function extensionality for uniform continuity) and address the other one in a future PR, for the sake of keeping the PR small.

martinescardo:
Let's merge this. Please remove the work-in-progress status so that I can do that.

ayberkt:
> Please remove the work-in-progress status so that I can do that.

I realized that a lot has changed after @brunorochapaiva and I worked on this, so I will have to sync it up with `master` for it to be mergeable. I’ll do that tomorrow.

martinescardo:
How is this going?

ayberkt:
Sorry @martinescardo, I looked at this after the last discussion and the revision it required turned out to be harder than I thought. I forgot to give you a status update.

I have now synced it up with the latest developments (a lot had changed after @brunorochapaiva and I implemented this refactoring). It should be ready to merge.

ayberkt:
@brunorochapaiva: Feel free to let me know if you have any suggestions. I’m not sure if I’ve used the machinery you’ve implemented in an optimal way.

By the way, we should also think about the question of fully removing function extensionality from the `InternalModUniCont` module.

brunorochapaiva:
I think it looks good. Looking through there are maybe one or two places where we could change it to use the hereditarily extensional equality explicitly, but it's a matter of taste whether this would simplify things.

martinescardo (approved):

PR #229, opened 2024-02-01
Delta complete poset

IanRay11:
Just adding more descriptive names to final results.


tomdjong:
(In the future, it's nicer if you pull `upstream/master` on your `master` and make a new branch so that the old commits that we've already merged don't show up again.)

PR #231, opened 2024-02-07
Add sigma closed subuniverses

awswan:
At the moment the library has definitions of subuniverse and reflective subuniverse. In order to get one of the 3 definitions of modalities given by Rijke, Shulman, Spitters, Modalities in HoTT, we have to add an extra condition to reflective subuniverse - that it is sigma closed. This patch adds the definition and a proof of the induction principle for sigma closed subuniverses.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Please add yourself the first time you contribute.
 
 * Alice Laroche
 * Andrew Sneap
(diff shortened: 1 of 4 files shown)


martinescardo (approved):
Thanks!

PR #233, opened 2024-02-07
Adding an alternative formulation of H-levels

IanRay11:
I also added my name to the read me file here. Hope that is okay.

### DIFF
--- a/source/UF/H-Levels.lagda
+++ b/source/UF/H-Levels.lagda
@@ -47,8 +47,7 @@ hlevel-relation-is-prop {𝓤} (succ n) X =
 
 \end{code}
 
-Some of the next results, left-inv and base, SHOULD be somewhere but I can't
-find them...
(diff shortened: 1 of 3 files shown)


source/UF/H-Levels.lagda
```diff
@@ -0,0 +1,163 @@
+Martin Escardo and Ian Ray, 06/02/2024
```
martinescardo, line 1: This should be "Ian Ray" only, I think.

martinescardo (changes requested):
Minimal change requested.

source/UF/H-Levels.lagda
IanRay11: I copied some of your proofs from the original file :)

IanRay11:
I am actually currently working on showing H n is of level (n + 1) then also, then we can merge. Unless you have more suggestions

PR #234, opened 2024-02-07
Sigma is singleton

IanRay11:


IanRay11:
I accidentally merged your master into this and it reopened. I'm not sure how to undo that... aopologies.

PR #235, opened 2024-02-08
Absoluteness of compactness

awswan:
This is a proof that compactness is upwards absolute with respect to reflective subuniverses. That is, if a type in a reflective subuniverse is compact relative to the internal logic of the subuniverse, then it is actually compact. This can be seen as a generalisation of the result that products of compact types indexed by a proposition are compact (see comments in file for more details).

This branch also adds the file to the index file in TypeTopology.

### DIFF
--- a/source/TypeTopology/AbsolutenessOfCompactness.lagda
+++ b/source/TypeTopology/AbsolutenessOfCompactness.lagda
@@ -291,7 +291,7 @@ extensionality and repleteness of the subuniverse.
 \end{code}
 
 The next two lemmas get quite technical. In both cases the ideas are
-simple, but we require constructing terms by by ○-induction or
+simple, but we require constructing terms by ○-induction or
 recursion. This requires proving that certain types are ○-modal, which
 requires some care with universe levels, as well as the application of
 several lemmas using function extensionality and repleteness of P.
--- a/source/TypeTopology/AbsolutenessOfCompactness.lagda
+++ b/source/TypeTopology/AbsolutenessOfCompactness.lagda
@@ -45,6 +45,7 @@ modality, and the weaker notion of reflective subuniverse suffices.
 https://doi.org/10.23638/LMCS-16(1:2)2020
 
 \begin{code}
+
 {-# OPTIONS --safe --without-K #-}
 open import MLTT.Spartan
 open import MLTT.Two-Properties
@@ -68,12 +69,14 @@ Modal.ReflectiveSubuniverse. In particular, we write ○ for the
 modality corresponding to the reflective subuniverse.
 
 \begin{code}
+
 module TypeTopology.AbsolutenessOfCompactness
  (P : subuniverse 𝓤 𝓥)
  (P-is-reflective : subuniverse-is-reflective P)
  where
 
 open import Modal.ReflectiveSubuniverse P P-is-reflective
+
 \end{code}
 
 We now give some statements related to compactness. We first consider
@@ -88,12 +91,14 @@ Unwinding all this, gives the following internal definition of
 compactness for the reflective subuniverse.
 
 \begin{code}
+
 is-internal-compact∙ : 𝓤 ̇ → 𝓤 ̇
 is-internal-compact∙ A =
  (F : A → ○ (Lift _ 𝟚))
  → Σ a₀ ꞉ A , (F a₀ ＝ η _ (lift 𝓤 ₁)
  → (a : A)
  → F a ＝ η _ (lift 𝓤 ₁))
+
 \end{code}
 
 It turns out that in addition to internal compactness, it's also
@@ -110,15 +115,18 @@ statement that A is compact. We will be able to show ○ (is-compact∙ A)
 internal compact.
 
 \begin{code}
+
 is-weak-internal-compact∙ : 𝓤 ̇ → 𝓤 ̇
 is-weak-internal-compact∙ A =
  (F : A → 𝟚)
    → Σ a₀ ꞉ A , (F a₀ ＝ ₁ → (a : A) → η _ (lift 𝓤 (F a)) ＝ η _ (lift 𝓤 ₁))
+
 \end{code}
 
 We check that weak internal compactness actually is weaker.
 
 \begin{code}
+
 internal-compact-implies-weak-internal-compact
  : (A : 𝓤 ̇ )
  → is-internal-compact∙ A
@@ -136,12 +144,14 @@ internal-compact-implies-weak-internal-compact A c F =
    : Σ a₀ ꞉ A , (F' a₀ ＝ η _ (lift 𝓤 ₁) → (a : A) → F' a ＝ η _ (lift 𝓤 ₁))
 
   weak-internal-instance = c F'
+
 \end{code}
 
 Note that we defined weak internal compactness so that it is also
 implied by compactness.
 
 \begin{code}
+
 compact-implies-weak-internal-compact
  : (A : 𝓤 ̇ )
  → is-compact∙ A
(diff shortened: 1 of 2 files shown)


source/TypeTopology/AbsolutenessOfCompactness.lagda
```diff
@@ -0,0 +1,403 @@
...
+We note that the result holds for all modalities with no further
+conditions and in particular the modality is not required to be lex,
+or to preserve 𝟘. For the main theorem, we don't even need a full
+modality, and the weaker notion of reflective subuniverse suffices.
+
+[1] Rijke, Shulman, Spitters, Modalities in homotopy type theory,
+https://doi.org/10.23638/LMCS-16(1:2)2020
+
+\begin{code}
+{-# OPTIONS --safe --without-K --exact-split --auto-inline #-}
```
martinescardo, line 48: The options `exact-split` and `auto-inline` are in the lib file and so are not needed. So are the other two, but I prefer to keep them for emphasis in all files.

source/TypeTopology/AbsolutenessOfCompactness.lagda
```diff
@@ -0,0 +1,403 @@
...
+open import MLTT.Spartan
+open import MLTT.Two-Properties
+
+open import UF.Base
```
martinescardo, line 52: I prefer the imports in alphabetical order if that's OK with you.

source/TypeTopology/AbsolutenessOfCompactness.lagda
```diff
@@ -0,0 +1,403 @@
...
+
+In this case we are thinking of reflective subuniverses as models of
+type theory sitting inside some larger "external" model of type
+theory. We will show that compactness is an upwards absolute
+notion. That is, if a type inside the reflective subuniverse is
+compact with respect to the internal logic of the subuniverse then it
+is compact viewed outside the subuniverse as just a type. The converse
+does not quite hold, so potentially there could be compact types where
+the internal statement of compactness is not true, and we don't get
+full absoluteness.
```
martinescardo, line 37: Interesting. Maybe you can add a TODO to think about a potential counter-example?

But I am a bit confused here. The type N of natural numbers is not compact in many models, including topological models, but the type (LPO -> N) is compact.

source/TypeTopology/AbsolutenessOfCompactness.lagda
```diff
@@ -0,0 +1,403 @@
...
+compact implies weak internal compact.
+
+This weaker notion will also be useful for making the connection with
+the results of TypeTopology.PropTychonoff clear. To do this, we will
+also look at the type obtained by simply applying the modality to the
+statement that A is compact. We will be able to show ○ (is-compact∙ A)
+→ is-weak-internal-compact∙, but not the same implication for just
+internal compact.
+
+\begin{code}
```
martinescardo, line 111: Oh, I see now, I think.

source/TypeTopology/AbsolutenessOfCompactness.lagda
```diff
@@ -0,0 +1,403 @@
...
+   where
+    t : Lift _ 𝟚 → A
+    t = 𝟚-cases a a₀ ∘ lower
+
+    t' : ○ (Lift _ 𝟚) → A
+    t' = ○-rec _ _ A-modal t
+
+    modal-zero-is-modal-one : η _ (lift _ ₀) ＝ η _ (lift _ ₁)
+    modal-zero-is-modal-one =
+     η _ (lift _ ₀) ＝⟨ ap (η _ ∘ lift _) (q ⁻¹) ⟩
```
martinescardo, line 219: Maybe align the equality symbols in this proof.

source/TypeTopology/AbsolutenessOfCompactness.lagda
```diff
@@ -0,0 +1,403 @@
...
+ module S =
+  Modal.SigmaClosedReflectiveSubuniverse
+   P P-is-reflective P-is-sigma-closed
+
+ fe = univalence-gives-funext ua
+ repleteness = univalence-implies-subuniverse-is-replete ua P
+\end{code}
+
+The next two lemmas get quite technical. In both cases the ideas are
+simple, but we require constructing terms by by ○-induction or
```
martinescardo, line 271: Double "by" ("by by").

martinescardo (changes requested):
This is really nice! I have minor comments, and a question that I think I more or less can answer now, but I will need to think a bit more about it.

You also need to add yourself to the list of contributors in the top `README.md` file. 

Thanks a lot for your contribution!

source/TypeTopology/AbsolutenessOfCompactness.lagda
martinescardo: So, in summary, keep only the first two options.

awswan:
I'll go ahead and make the minor changes first, and then answer the other questions.

source/TypeTopology/AbsolutenessOfCompactness.lagda
awswan: I don't have a clean Brouwerian counterexample that I can easily formalise in type theory and add, but I think this can happen in realizability models. In the effective topos the real numbers are precisely the computable reals and they are vacuously compact, because every function R -> 2 is constant. There's then a modality that "forces the halting set to be decidable." This adds non constant functions from the computable reals to 2 that prevents them from being compact in the subtopos, but I think the computable reals of the subtopos would still be compact in the effective topos, if that makes sense.

source/TypeTopology/AbsolutenessOfCompactness.lagda
```diff
@@ -0,0 +1,403 @@
...
+           _ _ _
+           (○-is-modal _)))
+\end{code}
+
+Finally, we can use the lemmas together with the main theorem to get a
+result which is closer to the statement of prop-tychonoff. This says ○
+"preserves compactness" in the sense that if ○ (A is compact), then
+(○ A) is compact.
+
+We note that the arguments in prop-tychonoff show that we can think of
```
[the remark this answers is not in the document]
awswan: I can probably rephrase this a bit to make it clearer. Is it the first paragraph or second paragraph you mean? The second paragraph is saying that when X is a proposition every family of types Y : X -> U is equal to the constant family X -> U that sends every element of x to \prod Y, and this was proved as part of prop-tychonoff.

source/TypeTopology/AbsolutenessOfCompactness.lagda
```diff
@@ -0,0 +1,403 @@
...
+the family of types Y : X → 𝓥 ̇  without loss of generality as a
+constant family. Aside from that and the assumption here of
+univalence, we can view the remainder of theorem prop-tychonoff as a
+special case of the one below. To see this, note that when X is a
+proposition the functor X → - is a modality, and so can derive
+is-compact∙ (X → A) from X → is-compact∙ A.
+
+\begin{code}
+ modalities-preserve-compact
```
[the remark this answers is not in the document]
awswan: I couldn't find an easy way to phrase the dependency using modalities. The open modality is just the exponential (-)^X so it doesn't have any dependency in the definition. There might be a way of rephrasing it by thinking of families of types Y : X -> U as elements of ○ U, but I think this is going to be more work, so I would prefer to leave for now. In particular, I think I would have to assume that the modality is lex, which means the formalisation would require adding lex modalities to the library.

source/TypeTopology/AbsolutenessOfCompactness.lagda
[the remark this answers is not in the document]
martinescardo: It's the sentence starting "We note that".

martinescardo (approved):
I'll leave the things that are currently unaddressed to a future pull request.

awswan:
I will do this, and some of the other comments, but I think now I'll have to come back to it next week.

PR #238, opened 2024-02-13
Absoluteness of compactness example

awswan:
Here is version 2 of the absoluteness of compactness. The main addition is using the open modality to derive another proof of prop-tychonoff. I had to do this as a separate file, since the modality is a module parameter to AbsolutenessOfCompactness, so in order to work with a specific modality (the open modality on a proposition P) it's necessary to pass it as a module parameter from a separate file.

I also added some explanation in the comments about an example from realizability of downwards absoluteness failing.

### DIFF
--- a/source/TypeTopology/AbsolutenessOfCompactnessExample.lagda
+++ b/source/TypeTopology/AbsolutenessOfCompactnessExample.lagda
@@ -21,7 +21,7 @@ open import UF.FunExt
 open import UF.PropIndexedPiSigma
 open import UF.Subsingletons
 
-module AbsolutenessOfCompactnessExample
+module TypeTopology.AbsolutenessOfCompactnessExample
  (fe : funext 𝓤 𝓤)
  (P : 𝓤 ̇ )
  (P-is-prop : is-prop P)
--- a/source/TypeTopology/AbsolutenessOfCompactnessExample.lagda
+++ b/source/TypeTopology/AbsolutenessOfCompactnessExample.lagda
@@ -1,0 +1,181 @@
+Andrew Swan, started 13th February 2024
+
+We demonstrate the upwards absoluteness of compactness, by using it to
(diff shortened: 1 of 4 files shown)


source/TypeTopology/AbsolutenessOfCompactnessExample.lagda
```diff
@@ -0,0 +1,181 @@
+Andrew Swan, started 13th February 2024
+
+We demonstrate the upwards absoluteness of compactness, by using it to
+give an alternative proof of the propositional Tychonoff theorem.
```
martinescardo, line 4: Maybe add `(proved in the module TypeTopology.PropTychonoff)`.

source/TypeTopology/AbsolutenessOfCompactnessExample.lagda
```diff
@@ -0,0 +1,181 @@
...
+non-dependent-prop-tychonoff = modalities-preserve-compact
+
+\end{code}
+
+In order to get a new proof of propositional Tychonoff, we need to
+show how to derive it from the non-dependent version.
+
+\begin{code}
+
+prop-tychonoff2
```
martinescardo, line 96: Minor stylistic suggestion: use subscript 2 instead?

source/TypeTopology/AbsolutenessOfCompactnessExample.lagda
```diff
@@ -0,0 +1,181 @@
...
+    fe open-is-replete P A A-modal
+
+\end{code}
+
+However, the fact that Π A is modal precisely tells us that the unit
+map of the modality Π A → (P → Π A) is an equivalence.
+
+Putting this together with the earlier proof that P → Π A is modal
+allows us to complete the proof of the theorem.
+
```
martinescardo, line 172: Nice!

PR #241, opened 2024-02-15
Define distributive lattices and their ideals

ayberkt:
Preparation for Stone duality on spectral locales.


martinescardo:
Let's discuss this in person next week. I feel that, for the purposes of Stone duality, it would be better to define distributive lattices purely equationally. (And then maybe prove some things that are part of your current definition as theorems, if necessary.)

ayberkt:
@martinescardo I have now changed the definition to the one in Stone Spaces and implemented the things in the previous definition as consequences of this definition.

source/Locales/DistributiveLattice/Definition.lagda
```diff
@@ -0,0 +1,340 @@
...
+ ∨-absorptive₃ x y = (y ∧ x) ∨ x ＝⟨ Ⅰ ⟩
+                     (x ∧ y) ∨ x ＝⟨ Ⅱ ⟩
+                     x           ∎
+                      where
+                       Ⅰ = ap (_∨ x) (∧-commutative y x)
+                       Ⅱ = ∨-absorptive₂ x y
+
+\end{code}
+
+Projection for the carrier set.
```
martinescardo, line 108: Projection -> Notation.

PR #259, opened 2024-04-12
Define frame isomorphisms

ayberkt:
This PR makes two changes.

1. It factors out the notions of frame homomorphism and continuous map of locales from the `Locales.Frame` module into their own modules.
1. Adds two variants of the notion of frame isomorphism together with a proof of their equivalence.


source/Locales/ContinuousMap/FrameIsomorphism-Definition.lagda
```diff
@@ -0,0 +1,245 @@
...
+
+Declare syntax for frame isomorphisms.
+
+\begin{code}
+
+Isomorphismᵣ-Syntax : Frame 𝓤 𝓥 𝓦 → Frame 𝓤' 𝓥' 𝓦 → 𝓤 ⊔ 𝓤' ⊔ 𝓥 ⊔ 𝓥' ⊔ 𝓦 ⁺  ̇
+Isomorphismᵣ-Syntax = FrameIsomorphisms.Isomorphismᵣ
+
+infix 0 Isomorphismᵣ-Syntax
+syntax Isomorphismᵣ-Syntax F G = F ≅f≅ G
```
martinescardo, line 243: Wouldn't `≅Fr≅` or even `≅⟨Fr⟩`  be better? I leave it up to you.

source/Locales/ContinuousMap/Homeomorphism-Definition.lagda
```diff
@@ -0,0 +1,61 @@
...
+
+Declare syntax for homeomorphisms.
+
+\begin{code}
+
+Homeomorphism-Syntax : Locale 𝓤 𝓥 𝓦 → Locale 𝓤' 𝓥' 𝓦 → 𝓤 ⊔ 𝓤' ⊔ 𝓥 ⊔ 𝓥' ⊔ 𝓦 ⁺  ̇
+Homeomorphism-Syntax = Homeomorphism
+
+infix 0 Homeomorphism-Syntax
+syntax Homeomorphism-Syntax X Y = X ≅c≅ Y
```
martinescardo, line 59: Why `c`?

martinescardo (approved):

source/Locales/ContinuousMap/Homeomorphism-Definition.lagda
ayberkt: I use the symbol `─c→` to denote continuous maps of locales, and the symbol `─f→` for frame homomorphisms.

I followed the same conventions for the notions of isomorphism given by these. I am open to suggestions though.

source/Locales/ContinuousMap/FrameIsomorphism-Definition.lagda
ayberkt: Yes I think that might actually be better. I think it is closer to conventions that I’ve seen in other parts of TypeTopology.

PR #260, opened 2024-04-18
Prove that isomorphic frames are equal

ayberkt:
This PR adds a proof that the notion of frame (as implemented in `frame-structure` in `Locales.Frame`) is a standard notion of structure.

Furthermore, it adds two corollaries of this fact:

1. Isomorphic frames are equal (using the structure identity principle).
2. “Transport lemma” for isomorphic frames: if a frame `F` satisfies some predicate `P : Frame (𝓤 ⁺) 𝓤 𝓤 → Ω 𝓣`  then any frame `G` with `F ≅ G` also satisfies `P`.


ayberkt:
This PR was closed because I changed the branch name from `stone-duality-2024-04-12` to `sip-for-frames`. I have created a PR (#261) for this branch under that new name.

PR #261, opened 2024-04-18
Prove that isomorphic frames are equal

ayberkt:
This PR adds a proof that the notion of frame (as implemented in `frame-structure` in `Locales.Frame`) is a standard notion of structure.

Furthermore, it adds two corollaries of this fact:

1. Isomorphic frames are equal (using the structure identity principle).
2. “Transport lemma” for isomorphic frames: if a frame `F` satisfies some predicate `P : Frame (𝓤 ⁺) 𝓤 𝓤 → Ω 𝓣`  then any frame `G` with `F ≅ G` also satisfies `P`.


source/Locales/ContinuousMap/Homeomorphism-Properties.lagda
```diff
@@ -0,0 +1,85 @@
...
+open AllCombinators pt fe
+open Locale
+
+\end{code}
+
+Homeomorphic locales are equal.
+
+\begin{code}
+
+homeomorphic-locales-are-equal : (X Y : Locale (𝓤 ⁺) 𝓤 𝓤) → X ≅c≅ Y → X ＝ Y
```
tomdjong, line 58: I would make the parameters of the locales as general as possible throughout this file.

ayberkt: I’ll do that in a separate PR in the future as to avoid spending any more time on this. As mentioned in PR #262, I wrote this down quickly to be able to prove a specific lemma about large and locally small spectral locales.

I will, however, write this down as a TODO for the reader who might be confused as to why this is done only for large and locally small locales.

source/Locales/SIP/FrameSIP.lagda
```diff
@@ -0,0 +1,341 @@
...
+open sip hiding (⟨_⟩)
+
+\end{code}
+
+We work in a module parameterized by two frame structures that we call `str₁`
+and `str₂`.
+
+\begin{code}
+
+module SIP-For-Frames {A : 𝓤 ⁺  ̇} (str₁ str₂ : frame-structure 𝓤 𝓤 A) where
```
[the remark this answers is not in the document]
ayberkt: I can but I don’t want to as this is not a priority.

source/Locales/ContinuousMap/Homeomorphism-Properties.lagda
tomdjong: Up to you — just think it's easier to generalize now than in the future.

tomdjong:
I'll leave further reviewing on this PR to @martinescardo.

source/Locales/ContinuousMap/Homeomorphism-Properties.lagda
ayberkt: That is a good point, but I’m a bit worried that it will require generalizing universes in a bunch of other places which will result in the PR becoming huge. At the moment, this PR achieves a well-defined goal in a self-contained way. I think it’s best if we merge this now, and move on to the generalization afterwards.

ayberkt:
Can you take a look at this when you get the chance @martinescardo?

PR #262, opened 2024-04-18
[Aborted] First step towards Stone duality for spectral locales

ayberkt:
**Do NOT merge before PR #261.**

---

This PR adds two things.

1. A new definition of spectral locale that we call `is-locale-of-spectra`.
1. A proof that `is-locale-of-spectra` implies `is-spectral-with-small-basis`.

Being able to prove (2) was my motivation for proving that isomorphic frames are equal (PR #261). It could have been done directly (I think), but I have come to the conclusion (after some discussions with @martinescardo, and also @tomdjong) that this will be the best way of doing it.

The motivation for doing it directly would be to see if univalence could have been avoided, but it seems like univalence is used for other purposes anyways so this doesn’t really make sense at the moment (but might turn out to be interesting for some investigations in the future).

CC: @IgorArrieta


tomdjong:
Can we make this a draft until #260 is merged?
Do you want me to review either/both of these?

ayberkt:
> Do you want me to review either/both of these?

@tomdjong Sure, if you have the time. I’ll make some small changes to PR #261 now and will send you a review request.

tomdjong:
I'm off next week — feel free to review & merge without me, or I can try to get round to it the week after :slightly_smiling_face:

ayberkt:
I’m aborting this PR and cherry-picking the relevant commits into PR #266.

PR #266, opened 2024-04-23
First step of Stone duality for spectral locales

ayberkt:
Cleaned up version of PR #262.


source/Locales/StoneDuality/ForSpectralLocales.lagda
```diff
@@ -0,0 +1,98 @@
...
+spec = locale-of-spectra
+
+\end{code}
+
+A locale `X` is said to be a _locale of spectra_ if it's homeomorphic to the
+locale of spectra of some distributive lattice `L `.
+
+\begin{code}
+
+is-locale-of-spectra : (X : Locale (𝓤 ⁺) 𝓤 𝓤) → 𝓤 ⁺  ̇
```
martinescardo, line 70: As we discussed in person, the correct definition is that a certain designated map is a homeomorphism, in this case the principal-ideal map. The current definition is not property, because e.g. X=1+1 will have two homeomorphisms.

martinescardo (changes requested):

ayberkt:
@martinescardo I have now truncated the definition as discussed. This PR should be ready to merge.

I will add the definition that specifies the equivalence in #270 instead of this one. The reason is that the two branches were supposed to be independent, but they ended up getting a bit tangled. The definition that we discussed needs some things from the other branch. So I will keep this PR self-contained and minimal, and continue the progress we discussed today in #270, after we merge this.

source/Locales/StoneDuality/ForSpectralLocales.lagda
```diff
@@ -0,0 +1,118 @@
...
+
+We denote by `spec L` the spectrum (the locale defined by the frame of ideals)
+of a distributive lattice `L`.
+
+\begin{code}
+
+open DefnOfFrameOfIdeal
+
+spec : DistributiveLattice 𝓤 → Locale (𝓤 ⁺) 𝓤 𝓤
+spec = locale-of-spectra
```
martinescardo, line 69: Didn't we discuss that there isn't such a thing as a locale of spectra and agree to change that?

martinescardo (changes requested):

source/Locales/StoneDuality/ForSpectralLocales.lagda
ayberkt: Yes, but this is not a change that I made in this PR. This is something already in `master` and I’m merely using it in this PR. I will change it as we discussed, just not in this PR. If I make that change here, the PR will involve many modules that is has nothing to do with. If you’re fine with that, I would be happy to make the change here.

source/Locales/StoneDuality/ForSpectralLocales.lagda
```diff
@@ -0,0 +1,118 @@
...
+date-started: 2024-04-12
+dates-updated: [2024-05-08]
+--------------------------------------------------------------------------------
+
+This module will eventually contain the proof Stone duality for spectral
+locales. It currently contains some preliminary results in this direction.
+
+\begin{code}[hide]
+
+{-# OPTIONS --safe --without-K --lossy-unification --exact-split --double-check --auto-inline #-}
```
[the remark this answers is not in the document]
martinescardo: Yes, this should be updated. Thanks.

PR #267, opened 2024-04-25
Transportation of distributive lattices along equivalences

ayberkt:


ayberkt:
This is ready to merge now @martinescardo.

source/Locales/DistributiveLattice/Isomorphism.lagda
```diff
@@ -0,0 +1,83 @@
+--------------------------------------------------------------------------------
```
martinescardo, line 1: Notice that a lattice isomorphism amounts to simply an equivalence that preserves and reflects order.

ayberkt: Noted down in #269.

PR #276, opened 2024-05-21
Take some first steps in investigating System F resizing as an axiom

ayberkt:
This PR adds a new module in which I summarize various discussions with @slspeight. Here is a short list of things contributed in the module:

- A special case of “System F resizing” that should be consistent.
- The proof that this implies `Ω¬¬`-resizing from `𝓤₁` to `𝓤₀`.
- A notion of `Σ`-resizing, which we’re not sure is consistent or not.
- A notion of `∃`-resizing, which is consistent because it is implied by propositional resizing.


source/UF/SystemFNotionOfResizing.lagda
```diff
@@ -0,0 +1,348 @@
...
+
+TODO: prove that this generalized form is inconsistent.
+
+
+The special case of this notion of resizing where we pick `𝓤 := 𝓤₀` and
+`𝓥 := 𝓤₁` should be consistent.
+
+\begin{code}
+
+System-F-Resizing : 𝓤₂  ̇
```
martinescardo, line 60: Should this really be called System F resizing? Is this the same form of impredicativity as in CoC?

martinescardo: In any case, if we agree that this is the same as in System F, it probably should be called System F impredicativity, I think.

source/UF/SystemFNotionOfResizing.lagda
```diff
@@ -0,0 +1,348 @@
...
+
+  β : ¬¬ (P⁺ holds) → ¬¬ (P holds)
+  β f p = f (p ∘ r)
+
+  γ : ¬¬-stable (P⁺ holds)
+  γ f = φ (β f) , ⋆
+
+\end{code}
+
+We now prove that propositional System F resizing implies Ω¬¬-resizing.
```
martinescardo, line 139: You already said this in line 97.

source/UF/SystemFNotionOfResizing.lagda
```diff
@@ -0,0 +1,348 @@
...
+One could also consider the propositional form of this notion of resizing.
+
+\begin{code}
+
+Propositional-System-F-Resizing : 𝓤₂  ̇
+Propositional-System-F-Resizing =
+ (A : 𝓤₁  ̇) →
+  (P : A → Ω 𝓤₀) →
+   (Ɐ x ꞉ A , P x) holds is 𝓤₀ small
+
```
martinescardo, line 75: If you choose `A` to be a proposition and `P` to be constantly `true`, then `(Ɐ x ꞉ A , P x) holds` is `A x 1` which is equivalent to A. Therefore, combining this with the statement in lines 88-89, we get that `Propositional-System-F-Resizing` is the same thing as propositional resizing.

source/UF/SystemFNotionOfResizing.lagda
martinescardo: This is because propositional resizing implies that, which we already know.

source/UF/SystemFNotionOfResizing.lagda
```diff
@@ -0,0 +1,348 @@
...
+\end{code}
+
+The version of this notion of resizing with truncation, which we denote
+∃-resizing, must be consistent as it is implied by propositional resizing.
+
+\begin{code}
+
+∃-Resizing : 𝓤₂  ̇
+∃-Resizing = (A : 𝓤₁  ̇) → (B : A → 𝓤₀  ̇) → (Ǝ x ꞉ A , B x) holds is 𝓤₀ small
+
```
martinescardo, line 343: Again consider `A` a proposition and `B x = 1` to conclude that this is equivalent to propositional resizing.

martinescardo (commented):
I think these results are still in premature form. You need to think more about this. I am not sure the propositional versions are interesting.

source/UF/SystemFNotionOfResizing.lagda
ayberkt: > If you choose `A` to be a proposition and `P` to be constantly `true`, then `(Ɐ x ꞉ A , P x) holds` is `A x 1`

That would be `A ⇒ 𝟙`, which is just `𝟙`, and not `A × 𝟙`.

In fact, this implication would mean that `System-F-Resizing` is equivalent to `propositional-resizing`, which cannot be the case since the former is validated and the latter is refuted in the cubical assemblies model. I learned this from @slspeight.

So I think `Propositional-System-F-Resizing` is actually a sensible thing to consider.

source/UF/SystemFNotionOfResizing.lagda
ayberkt: Right, this one _is_ equivalent to propositional resizing.

Issue #178, opened 2023-09-07
Add a proof of Zorn’s Lemma


ayberkt:
As mentioned by @martinescardo on Discord.

ayberkt:
Closing this as it has been completed by @keltono.

PR #281, opened 2024-06-01
Proof of Zorn's lemma

keltono:
A proof that the Axiom of Choice implies Zorn's Lemma.
Should close or at least partially address issue #178 . Does not provide a proof that Zorn's Lemma (+ LEM) implies the Axiom of choice.
This is my first time contributing here (or really doing a substantial formalization project...), so let me know if there are any issues with choice of definitions/style!

### DIFF
--- a/source/OrderedTypes/ZornsLemma.lagda
+++ b/source/OrderedTypes/ZornsLemma.lagda
@@ -1,38 +1,38 @@
 Kelton OBrien, 31 May 2024
 
 A proof that the Axiom of Choice implies Zorn's Lemma, as well as relevant definitions.
+
 \begin{code}
 
 {-# OPTIONS --safe --without-K --exact-split #-}
 
 open import MLTT.Spartan
+open import Ordinals.Equivalence
 open import Ordinals.Notions
 open import Ordinals.Type
+open import Ordinals.Underlying
+open import UF.Base
+open import UF.Choice
+open import UF.ClassicalLogic
+open import UF.DiscreteAndSeparated
+open import UF.Embeddings
+open import UF.Equiv
+open import UF.EquivalenceExamples
+open import UF.FunExt
+open import UF.Logic
+open import UF.Powerset-MultiUniverse
+open import UF.PropTrunc
 open import UF.Sets
 open import UF.Size
-open import UF.Embeddings
-open import UF.Powerset-MultiUniverse
-open import UF.SubtypeClassifier
-open import UF.DiscreteAndSeparated
-open import UF.PropTrunc
-open import UF.ClassicalLogic
-open import UF.Univalence
-open import UF.FunExt
-open import UF.UA-FunExt
-open import Ordinals.Underlying
-open import Ordinals.Equivalence
-open import UF.Logic
-open import UF.Base
-open import UF.Equiv
 open import UF.Subsingletons
-open import UF.Choice
 open import UF.Subsingletons-FunExt
-open import UF.EquivalenceExamples
+open import UF.SubtypeClassifier
+open import UF.UA-FunExt
+open import UF.Univalence
 open import UF.UniverseEmbedding
 
 module OrderedTypes.ZornsLemma
- {𝓤 : Universe}
- {𝓣 : Universe}
+ {𝓤 𝓣 : Universe}
  (pt : propositional-truncations-exist)
  (ua : Univalence)
  {X : 𝓤 ̇}
@@ -40,8 +40,8 @@ module OrderedTypes.ZornsLemma
   where
 
 open PropositionalTruncation pt
-open import Ordinals.OrdinalOfOrdinals ua
 open import Ordinals.BuraliForti ua
+open import Ordinals.OrdinalOfOrdinals ua
 open import UF.ImageAndSurjection pt
 
 fe : FunExt
@@ -53,25 +53,30 @@ pe' 𝓤 = pe {𝓤}
 fe' : Fun-Ext
 fe' {𝓤} {𝓥} = fe 𝓤 𝓥
 
-open UF.Choice.choice-functions pt pe' fe
+open UF.Choice.ExcludedMiddle pt fe
 open UF.Choice.Univalent-Choice fe pt
-open import Ordinals.Arithmetic fe
+open UF.Choice.choice-functions pt pe' fe
+open UF.Logic.Existential pt
+open UF.Logic.Universal fe'
 open import OrderedTypes.Poset fe'
+open import Ordinals.Arithmetic fe
 open import Ordinals.WellOrderingTaboo fe' pe
+open inhabited-subsets pt
+
 open PosetAxioms
 open ClassicalWellOrder pt
-open UF.Logic.AllCombinators pt fe'
-open UF.Choice.ExcludedMiddle pt fe
-open inhabited-subsets pt
 
 _⋘_ : X → X → (𝓤 ⊔ 𝓣) ̇
 a ⋘ b =  a ≪ b × (a ≠ b)
 
 is-chain : {𝓥 : Universe} → (Y : 𝓟 {𝓥} X ) →  (𝓤 ⊔ 𝓣 ⊔ 𝓥) ̇
-is-chain Y = (x y : X) → x ∈ Y → y ∈ Y → ∥ (x ≪ y + y ≪ x) ∥
+is-chain Y = (x y : X) → x ∈ Y → y ∈ Y → (x ≪ y ∨ y ≪ x)
 
 has-maximal-element-strong :  (𝓤 ⊔ 𝓣) ̇
 has-maximal-element-strong = Σ x ꞉ X , ((y : X) → x ≪ y → x ＝ y)
+
+has-maximal-element : (𝓤 ⊔ 𝓣) ̇
+has-maximal-element = ∃ x ꞉ X , ((y : X) → x ≪ y → x ＝ y)
 
 all-chains-have-upper-bound : {𝓥 : Universe} → (𝓤 ⊔ 𝓣 ⊔ (𝓥 ⁺))  ̇
 all-chains-have-upper-bound {𝓥} =
@@ -79,10 +84,14 @@ all-chains-have-upper-bound {𝓥} =
 
 \end{code}
 
-The following is not a direct formalization of any extant proof,
-but follows the same proof idea as the standard transfinite-induction-based
+The following is a formalization ofthe standard transfinite-induction-based
 proof of Zorn's lemma. The closest informal analog to the proof can be found
-here: https://math.stackexchange.com/a/2889205 .
+here: https://math.stackexchange.com/a/2889205.
+The primary differences between this proof and that proof are just ones of
+how sets are treated (notice the lack of a particular poset P),
+additional detail in things that are glossed over in the formal proof,
+(e.g the proof that the chain allfbs is a chain is not given).
+The proof idea is the same.
 
 We do a proof by contradiction (which is why we have that double negation),
 where we assume that even though all chains have a least upper bound,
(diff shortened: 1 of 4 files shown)


ayberkt:
Thanks @keltono! This is something I've wanted to see in TypeTopology for a long time and I'm happy to see that someone's finally done it.

I'll try to review it as soon as possible, hopefully this weekend.

martinescardo:
Thanks a lot! I will review this.

martinescardo:
Looks good! I will do a more thorough review later. For the moment, you should add two more things to your pull request: (1) Your name at the top README.md file, and (2) An intry to the index file of the directory your new file is in.

ayberkt:
@keltono May I also suggest improvements on cosmetic issue that I have noticed? I will do a more proper review later.

1. Instead of `∥ (x ≪ y + y ≪ x) ∥` can you use `∨` from `UF.Logic`?
2. Can you make sure that there are blank lines after `\begin{code}` and before `\end{code}`?

keltono:
Alright, I addressed everything except the point about ∨. 
I have some thoughts about that:
posets don't  (necessarily) yield an Omega type, so that information would have to get carried around in proof when it's not really needed, as Zorn's lemma already assumes that ≪ is a poset (and as such x≪y is a proposition) .
WIth that  said, it seems like ∨ from `PropTrunc` would be more appropriate, but I seem to be running into an agda parsing bug where agda can't parse   `x ≪ y ∨ y ≪ x` since it's ambiguous if it's  `UF.PropTrunc.∨` or if it's `UF.Logic.∨`, even though `UF.Logic.∨` is not in scope (since I don't  open/import `Disjunction`), which I haven't been able to resolve by using the fully qualified name. Not sure what's going on there. I have been able to solve this by hiding the `PropTrunc` ∨ ---which allows me to use the `Logic  ∨` after importing Disjunction ---  but hiding the `Logic` ∨ doesn't work (since it's not in scope!), so that would force me to use the `Logic` ∨, which I want to avoid for the reason I mentioned above. There might be some way around this I don't know, I'm not an agda module system expert :). 

In any case, I think it's probably fine as is, but you all disagree then it would not be a problem to just switch to the UF.Logic version --- it just feels a little redundant to me.

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
+Kelton OBrien, 31 May 2024
+
+A proof that the Axiom of Choice implies Zorn's Lemma, as well as relevant definitions.
+\begin{code}
```
martinescardo, line 4: I prefer to have blank lines before and after `\begin{code}` and `\end{code}`.

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
...
+
+{-# OPTIONS --safe --without-K --exact-split #-}
+
+open import MLTT.Spartan
```
martinescardo, line 8: Maybe sort the imports? You can do this by marking the block in `emacs` and then running `sort-lines`.

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
...
+open import UF.Base
+open import UF.Equiv
+open import UF.Subsingletons
+open import UF.Choice
+open import UF.Subsingletons-FunExt
+open import UF.EquivalenceExamples
+open import UF.UniverseEmbedding
+
+module OrderedTypes.ZornsLemma
+ {𝓤 : Universe}
```
martinescardo, line 34: You could put the two universes in a single declaration.

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
...
+is found below, and takes in more reasonable inputs.
+
+\begin{code}
+
+choice-function-gives-not-not-zorns-lemma :
+ Excluded-Middle →
+ poset-axioms (_≪_) →
+ all-chains-have-upper-bound  →
+ (Σ ε ꞉ (𝓟 {𝓣 ⊔ 𝓤} X → X) , ((A : 𝓟 X) → is-inhabited A → ε A ∈ A)) →
+ ¬¬ has-maximal-element-strong
```
martinescardo, line 108: Given that you are assuming excluded middle, you could use `DNE` and `EM-gives-DNE` to conclude `has-maximal-element-strong` without `¬¬`.

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
...
+    dn : {𝓥 : Universe } → DNE 𝓥
+    dn {𝓥} = EM-gives-DNE lem
+
+    eq-is-¬¬-stable : {x : X} {y : X } →  ¬¬-stable (x ＝ y)
+    eq-is-¬¬-stable {x} {y} ¬¬x=y = dn (x ＝ y) (X-is-set) ¬¬x=y
+
+    ≪-is-¬¬-stable : {x : X} {y : X} → ¬¬-stable (x ≪ y)
+    ≪-is-¬¬-stable {x} {y} ¬¬ll = dn (x ≪ y) (≪-prop x y) ¬¬ll
+
+    ¬¬Σ→∃ : {𝓤 : Universe} {A : X → 𝓤  ̇} → ¬ ¬ (Σ x ꞉ X , A x ) → (∃ x ꞉ X , A x)
```
martinescardo, line 127: Maybe this could go to `UF.ClassicalLogic` or somewhere else, given that it is a fact of general use when one is doing classical logic.

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
...
+        allfb-has-ub = chains allfb allfb-is-chain
+
+        ub : X
+        ub = pr₁ allfb-has-ub
+
+        ub-is-ub : ∀ y → y ∈ allfb → y ≪ ub
+        ub-is-ub = pr₂ allfb-has-ub
+
+        ub-is-strict' : (Σ q ꞉ X , (¬ ( ub ≪ q → ub ＝ q)))
+                         →  Σ z ꞉ X , (∀ y → y ∈ allfb → y ⋘ z)
```
martinescardo, line 217: Align ":" and "→ " here and elsewhere as appropriate.

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
...
+      less : (a b : Ordinal 𝓤) → a ⊲ b → f a ≠ f b
+      less a b a<b = pr₂ (f-preserves-order b a a<b)
+
+      more : (a b : Ordinal 𝓤) → b ⊲ a → f a ≠ f b
+      more a b b<a = ≠-sym (pr₂ (f-preserves-order a b b<a))
+
+      equal : (a b : Ordinal 𝓤) → a ≠ b → a ＝ b → f a ≠ f b
+      equal a b a≠b a=b = unique-from-𝟘 (a≠b a=b)
+
+    f-is-left-cancellable : {a : Ordinal 𝓤} → {b : Ordinal 𝓤} → f a ＝ f b → a ＝ b
```
martinescardo, line 250: Maybe write just `{a b : Ordinal 𝓤}`.

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
...
+    f-is-left-cancellable {a} {b} p =
+     dn (a ＝ b) (the-type-of-ordinals-is-a-set (ua 𝓤) fe')
+                 ((contrapositive (f-is-injective a b)) (¬¬-intro p))
+
+    f-is-embedding : is-embedding f
+    f-is-embedding = lc-maps-into-sets-are-embeddings f f-is-left-cancellable X-is-set
+
+    X' : 𝓤 ⁺ ̇
+
```
martinescardo, line 259: Remove blank line.

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
...
+ → poset-axioms (_≪_)
+ → all-chains-have-upper-bound
+ → (Σ ε ꞉ (𝓟 {𝓣 ⊔ 𝓤} X → X) , ((A : 𝓟 X) → is-inhabited A → ε A ∈ A))
+ → has-maximal-element
+
+choice-function-gives-zorns-lemma lem axioms chains cf = ¬¬Σ→∃ ¬¬max
+ where
+  dn : {𝓥 : Universe } → DNE 𝓥
+  dn {𝓥} = EM-gives-DNE lem
+  ¬¬Σ→∃ : {𝓤 : Universe} {A : X → 𝓤 ̇} → ¬ ¬ (Σ x ꞉ X , A x ) → (∃ x ꞉ X , A x)
```
martinescardo, line 314: Didn't we have this already above?

martinescardo: There is an abbreviation `¬¬` for `¬ ¬` that could be used here.

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
...
+  ¬¬Σ→∃ {𝓤} {A} inside = dn _ ∥∥-is-prop (¬¬-functor ∣_∣ inside)
+  ¬¬max : ¬ ¬ has-maximal-element-strong
```
martinescardo, line 316: And here.

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
...
+  ¬¬max = (choice-function-gives-not-not-zorns-lemma lem axioms chains cf)
+
+axiom-of-choice-implies-zorns-lemma
+ : Axiom-of-Choice
+ → poset-axioms (_≪_)
+ → (all-chains-have-upper-bound → has-maximal-element)
```
martinescardo, line 322: Do you really need the brackets here?

source/OrderedTypes/ZornsLemma.lagda
```diff
@@ -0,0 +1,375 @@
...
+
+  choice-function : ∥ X ∥ → ∃ ε ꞉ (𝓟 X → X) , ((A : 𝓟 X) → is-inhabited A → ε A ∈ A)
+  choice-function isX =  ∥∥-functor lower-cf (lifted-cf (∥∥-functor (lift (𝓤 ⊔ 𝓣)) isX))
+
+  I'
+   : all-chains-have-upper-bound
+   → ∃ ε ꞉ (𝓟 X → X) , ((A : 𝓟 X) → is-inhabited A → ε A ∈ A)
+   → has-maximal-element
+  I' chains = ∥∥-rec (∃-is-prop)
+          (choice-function-gives-zorns-lemma em (X-is-set , axioms-rest) chains)
```
martinescardo, line 360: Indentation could be better here.

martinescardo (changes requested):
I've made only stylistic comments.
I still need to fully understand this particular proof you are offering. But it looks very nice. I wonder if you could discuss a complete informal outline of the proof before the formal proof? In any case, thanks again for writing this! We definitely need Zorn's Lemma in TypeTopology.

martinescardo:
> I wonder if you could discuss a complete informal outline of the proof before the formal proof?

Maybe this is not needed if you only explain the main differences with the MO answer.

ayberkt:
> WIth that said, it seems like ∨ from PropTrunc would be more appropriate

Yes sorry, I didn’t mean specifically the one from `UF.Logic`. Either the one from `PropTrunc` or `UF.Logic` is fine. I can try to help you with the parsing problem.

keltono:
> Maybe this is not needed if you only explain the main differences with the MO answer.
After giving it a thorough look through, I think that I was  perhaps overselling the differences. The idea is pretty much the same between the two, modulo some particular things about how sets are dealt with in type theory. I changed the comment accordingly.

I was also able to figure out the parsing issue :).  I think all of the comments should have been addressed, let me know if there are further points.

source/OrderedTypes/ZornsLemma.lagda
keltono: I think to get that I would have to weaken `maximal-element-strong` to `maximal-element` (like I do later) since I don't think maximal-element-strong is a proposition. The other problem I had there as needing `¬ has-maximal-element-strong` to get the contradiction to begin with --- right now I'm taking that as an argument to the top-level function which I think looks a bit better, since otherwise all of the terms need to be parametrized by `¬ has-maximal-element strong` but the change I just pushed changes it to make the term that's being constructed of type `¬ has-maximal-element strong →   𝟘`.

source/OrderedTypes/ZornsLemma.lagda
keltono: I agree but I'm not sure how I would change it. Any further thoughts?

martinescardo:
Thanks again!

source/OrderedTypes/ZornsLemma.lagda
martinescardo: I usually do this like this:
```
∥∥-rec 
 (∃-is-prop)
 (choice-function-gives-zorns-lemma em (X-is-set , axioms-rest) chains)
``` 
In general,
```
some-function
 argument1
 argument2
 ...
```
In more complicated situations,
```
some-function
 argument1
 (some-other-function
    argumenta
    argumentb)
 argument3
```

martinescardo: You can fix this in a future pull request.

PR #282, opened 2024-06-02
The large dcpo of ordinals

tomdjong:


source/DomainTheory/Examples/Ordinals.lagda
```diff
@@ -0,0 +1,172 @@
...
+\end{code}
+
+Unlike many other examples, such as the dpcos in the Scott model of PCF, or
+Scott's D∞, the dpco of ordinals, while algebraic, does not have a small
+(compact) basis. If it did, we could take the join of all the basis elements to
+obtain a greatest ordinal, which does not exist.
+
+\begin{code}
+
+Ordinals-DCPO-has-no-small-basis : ¬ (has-unspecified-small-basis Ordinals-DCPO)
```
ayberkt, line 158: Nice to have this counterexample formalized. I have yet to find a frame that does not have a small basis.

UlrikBuchholtz: > Nice to have this counterexample formalized. I have yet to find a frame that does not have a small basis.

What about not-not stable propositions (without propositional resizing)?

tomdjong:
Maybe I'm missing something, but presumably the booleans form a small
basis, just like for the type of all (small) propositions?

source/DomainTheory/Examples/Ordinals.lagda
```diff
@@ -0,0 +1,172 @@
...
+  (greatest-element-if-sup-complete-with-small-basis
+    Ordinals-DCPO
+    Ordinals-DCPO-is-sup-complete
+    h)
+
+Ordinals-DCPO-has-no-small-compact-basis :
+ ¬ (has-unspecified-small-compact-basis Ordinals-DCPO)
+Ordinals-DCPO-has-no-small-compact-basis h =
+ Ordinals-DCPO-has-no-small-basis
+  (∥∥-functor (λ (B , β , scb) → B , β , compact-basis-is-basis _ β scb) h)
```
martinescardo, line 170: Therefore it doesn't have a specified basis either. Maybe add this remark?

martinescardo (approved):
Nice!

source/DomainTheory/Examples/Ordinals.lagda
ayberkt: The type of not-not-stable propositions would not work since the Booleans form a basis for them, as they do for $\Omega$. However, there is a resizing taboo in this direction, which might be what you had in mind @UlrikBuchholtz.

In [our preprint](https://arxiv.org/abs/2402.03134) on the patch topology with @martinescardo and @IgorArrieta, we have two equivalent definitions of the notion of spectral locale:

1. The one in Definition 4.5 which says that the type of compact opens forms a small basis closed under binary meets.
2. The one that says there merely exists a small spectral basis (defined in Definition 4.9) for the locale in consideration.

In Theorem 4.17, it is shown that they coincide, but this crucially relies on the locale in consideration being compact.

For zero-dimensional locales (not necessarily Stone), one can consider the analogues of these definitions and ask if they coincide.

They don’t coincide and a counterexample to this is $\Omega_{\neg\neg}$; its type of clopens is not small but there exists a small basis consisting of clopens. By going from $\Omega$ to $\Omega_{\neg\neg}$ (the smallest dense sublocale of $\Omega$), the size of the clopens increases from just $2$ to $\Omega$-many because everything in $\Omega_{\neg\neg}$ is a clopen. So if the two definitions coincide on it, then $\Omega_{\neg\neg}$-resizing holds.

This counterexample was found by @IgorArrieta when I asked him I was looking for such a counterexample.

martinescardo: Interesting. But is there an example where provably there isn't a small base?

ayberkt: @martinescardo No, there is no example where the nonexistence of a small base is provable. There is not even an example where the existence of a small basis gives a resizing taboo.

PR #286, opened 2024-06-07
The lifting of a large proposition as an algebraic dcpo

tomdjong:
...has a small compact basis iff the proposition can be resized.

cc @ayberkt


martinescardo:
Wonderful! I don't have anything to suggest.

PR #289, opened 2024-06-23
If the flat natural numbers are ω-complete, then LPO holds

tomdjong:


martinescardo:
Nice! I have one minor "complaint". In `CoNatutorals/Sharp` we define
```
ℕ⊥ = 𝓛 ℕ
```
and call this the flat natural numbers. 

Now you have something else called the same, with the same notation.

At least a warning would be needed, saying that one of them is the one that works constructively, and the other one is the classical definition (and that they are equivalent classically).

(I forget what we call things in our papers and in your thesis.)

Everything else in the PR is fine.

tomdjong:
Thanks @martinescardo! I added a warning.

source/DomainTheory/Taboos/ClassicalLiftingOfNaturalNumbers.lagda
```diff
@@ -0,0 +1,232 @@
...
+
+The poset is defined as follows. Notice that the set of natural numbers is
+ordered discretely.
+
+NB: In this file, and this file only, we denote this poset by ℕ⊥. In
+CoNaturals/Sharp this notation is reserved for the lifting of the natural
+numbers which constructively *does* yield a directed complete
+poset. Classically, the constructions are equivalent, as formalized in
+Lifting.Miscelanea-PropExt-FunExt.
+
```
martinescardo, line 47: Good. Maybe you can add that this the case only classically, and that it is the purpose of this file to prove this.

martinescardo (approved):
Thanks again. I think the proposed solution to my initial "complaint" is fine. I just suggested one more thing that could be said. Although it is already implicitly said, I think it would be good to say it explicitly for emphasis.

source/DomainTheory/Taboos/ClassicalLiftingOfNaturalNumbers.lagda
martinescardo: Let me know if you want me to merge this as it is, or if you would be happy to add the additional remark.

martinescardo (approved):
Perfect! Thanks.

PR #299, opened 2024-07-31
Fundamental theorem of identity types

tomdjong:
Your comments are welcome @martinescardo.


martinescardo:
For the record, this was deleted because it is already present in `UF.Yoneda`.

PR #300, opened 2024-07-31
Answering Multiset-Addendum's question

alicelaroche:
Here is the file that answer the question asked in Multiset-Addendum. Comments are welcome, the indentation might not be strange in some places.

### DIFF
--- a/source/gist/multiset-addendum-question.lagda
+++ b/source/gist/multiset-addendum-question.lagda
@@ -37,16 +37,16 @@ open import Iterative.Multisets 𝓤
 open import Iterative.Multisets-Addendum ua 𝓤
 
 swap-Idtofun : {X Y : 𝓤 ̇ } {Z : 𝓥 ̇ } → {f : X → Z} {g : Y → Z}
-      → (p : Y ＝ X)
-      → f ∘ Idtofun p ＝ g
-      → f ＝ g ∘ Idtofun⁻¹ p
+             → (p : Y ＝ X)
+             → f ∘ Idtofun p ＝ g
+             → f ＝ g ∘ Idtofun⁻¹ p
 swap-Idtofun  refl refl = refl
 
 Question𝟘 :
  ¬ (Σ Πᴹ ꞉ ((𝟘 {𝓤} → 𝕄) → 𝕄)
-           , ((A : 𝟘 → 𝕄) → Πᴹ A ＝ ssup
-                                     (Π x ꞉ 𝟘 , 𝕄-root (A x))
-                                     (λ g → Πᴹ (λ x → 𝕄-forest (A x) (g x)))))
+         , ((A : 𝟘 → 𝕄) → Πᴹ A ＝ ssup
+                                  (Π x ꞉ 𝟘 , 𝕄-root (A x))
+                                  (λ g → Πᴹ (λ x → 𝕄-forest (A x) (g x)))))
 Question𝟘 (Πᴹ , eq) = recurs A (Πᴹ A) (eq A)
  where
   A : 𝟘 → 𝕄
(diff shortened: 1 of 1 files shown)


source/gist/multiset-addendum-question.lagda
```diff
@@ -0,0 +1,203 @@
...
+ data _<_ : (X → 𝕄) → (X → 𝕄) → (𝓤 ⁺) ̇  where
+  smaller : {f g : X → 𝕄} → ((x : X) → f x ⁅ g x) → f < g
+
+ open import Ordinals.Notions _<_
+
+ <-is-well-founded : X → is-well-founded
+ <-is-well-founded x f = acc (rec' x f (f x) refl)
+  where 
+   rec' : (x : X) (f : X → 𝕄) (m : 𝕄) → m ＝ f x
+      → (g : X → 𝕄) → g < f
```
martinescardo, line 87: One more space in line with the above comment.

source/gist/multiset-addendum-question.lagda
```diff
@@ -0,0 +1,203 @@
...
+        → (acc₁ acc₂ : is-accessible A) → Πᴹ' A acc₁ ≃ᴹ Πᴹ' A acc₂
+    III A rec acc₁ acc₂ = transport₂ _≃ᴹ_ (II A acc₁ ⁻¹) (II A acc₂ ⁻¹)
+                           ((≃-refl _)
+                           , λ g → rec (λ x → 𝕄-forest (A x) (g x))
+                                       (smaller λ y → (g y) , refl)
+                                       (prev acc₁ _ _)
+                                       (prev acc₂ _ _))
+
+    IV : (A : X → 𝕄) → (acc₁ acc₂ : is-accessible A)
+        → Πᴹ' A acc₁ ≃ᴹ Πᴹ' A acc₂
```
martinescardo, line 146: Align with colon.

martinescardo (approved):
The indentation is good. I've just made a few pedantic suggestions. I will accept it as it is, but I suggest you make these minor changes when you have time. Thanks for your contribution.

source/gist/multiset-addendum-question.lagda
```diff
@@ -0,0 +1,203 @@
...
+ ¬ (Σ Πᴹ ꞉ ((𝟘 {𝓤} → 𝕄) → 𝕄)
+         , ((A : 𝟘 → 𝕄) → Πᴹ A ＝ ssup
+                                  (Π x ꞉ 𝟘 , 𝕄-root (A x))
+                                  (λ g → Πᴹ (λ x → 𝕄-forest (A x) (g x)))))
+Question𝟘 (Πᴹ , eq) = recurs A (Πᴹ A) (eq A)
+ where
+  A : 𝟘 → 𝕄
+  A x = 𝟘-elim x
+
+  recurs : (A : 𝟘 → 𝕄) → (x : 𝕄)
```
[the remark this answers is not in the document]
ayberkt: @martinescardo This reminds me that it might be a good idea to add a bit more details to the clause on this in the contribution guidelines.

When the type declaration `f : X → Y → Z` is broken into separate lines, the convention is to indent it like:

```agda
f : X
  → Y
  → Z
```

as to avoid arrows on the same line.

What do we do if we have something like the function below?

```agda
f : (x x' : X) (y : Y(x)) → Z(x, x', y)
```

Is it better to write this as

```agda
f : (x x' : X)
  → (y : Y(x))
  → (z : Z(x, x', y))
```

or as

```agda
f : (x x' : X) (y : Y(x))
  → Z(x, x', y)
```

as to avoid multiple quantifications on the same line?

I sometimes have a hard time deciding between the two. If either of these is a more standard convention, I would be happy to write it down for future reference.

PR #301, opened 2024-08-01
Add alternative characterizations of compactness and clean up

ayberkt:
This PR
  * organizes the modules on compactness and places them under a new `Compactness` subdirectory, and
  * adds a proof that the following definitions of compactness are all equivalent:
     1. For every directed family $S$,  if $U \le \bigvee S$ then $\exists i : I.\ U \le S_i$.
     2. For every family $S$, if  $U \le \bigvee S$ then there is a Kuratowski-finite subfamily $F \subseteq S$ with $U \le \bigvee F$.
     3. For every family $S$, if $U = \bigvee S$, then there is a Kuratowski-finite subfamily $F \subseteq S$ with $U = \bigvee F$.

This was motivated by the presentation of compactness in the thesis.


source/Locales/Compactness/Properties.lagda
```diff
@@ -0,0 +1,716 @@
...
+SubFam {𝓤} {A} {𝓦} (I , α) = Σ J ꞉ 𝓦  ̇ , (J → I)
+
+\end{code}
+
+Tiny lemma recording the fact that nothing is a member of the empty list.
+
+\begin{code}
+
+not-in-empty-list : {A : 𝓤  ̇} {x : A} → ¬ ∥ member x [] ∥
+not-in-empty-list = ∥∥-rec 𝟘-is-prop (λ ())
```
martinescardo, line 71: Several comments here. (1) This should probably go to MLTT.List. (2) In general, ¬ A is equivalent to ¬ ∥ A ∥ for any type A. (3) We should never use λ () - we should instead use the eliminator of the empty type.

source/Locales/Compactness/Properties.lagda
```diff
@@ -0,0 +1,716 @@
...
+nth {_} {X} (_ ∷ xs) (inl n) = x , ∥∥-functor in-tail (pr₂ IH)
+ where
+  IH : Σ x ꞉ X , ∥ member x xs ∥
+  IH = nth xs n
+
+  x : X
+  x = pr₁ IH
+
+nth-is-surjection : {X : 𝓤  ̇} (xs : List X) → is-surjection (nth xs)
+nth-is-surjection []       (y , μ) = ∥∥-rec ∃-is-prop (λ ()) μ
```
martinescardo, line 91: Again.

source/Locales/Compactness/Properties.lagda
```diff
@@ -0,0 +1,716 @@
...
+
+list-members-is-Kuratowski-finite : {X : 𝓤  ̇}
+                                  → (xs : List X)
+                                  → is-Kuratowski-finite (Σ x ꞉ X , ∥ member x xs ∥)
+list-members-is-Kuratowski-finite {𝓤} {A} xs =
+ ∣ length xs , nth xs , nth-is-surjection xs ∣
+
+\end{code}
+
+TODO: The function `nth` above should be placed in a more appropriate module.
```
martinescardo, line 113: MLTT.List.

source/Locales/Compactness/Properties.lagda
```diff
@@ -0,0 +1,716 @@
...
+     × (U ≤ (⋁[ 𝒪 X ] ⁅  S [ h j ] ∣ j ∶ J ⁆)) holds)
+  where
+   open PosetNotation (poset-of (𝒪 X))
+
+\end{code}
+
+This is much closer to the “every cover has a finite subcover” definition from
+point-set topology.
+
+It is easy to show that this implies the standdard definition of compactness,
```
martinescardo, line 136: Typo: "Standdard".

source/Locales/Compactness/Properties.lagda
```diff
@@ -0,0 +1,716 @@
...
+but we need a bit of preparation first.
+
+Given a family `S`, we denote by `family-of-lists S` the family of families
+of lists of `S`.
+
+\begin{code}
+
+module Some-Lemmas-On-Directification (F : Frame 𝓤 𝓥 𝓦) where
```
martinescardo, line 144: Why not everything in lower case? Either use hyphens (preferred) or camel case.

source/Locales/Compactness/Properties.lagda
```diff
@@ -0,0 +1,716 @@
...
+    ‡ = ∨[ F ]-least ‡₁ ‡₂
+     where
+      ‡₁ : (S [ i ] ≤ directify₂ S [ i ∷ is ]) holds
+      ‡₁ = ⋁[ F ]-upper (family-of-lists S [ i ∷ is ]) (i , ∣ in-head ∣)
+
+      ‡₂ : (directify₂ S [ is ] ≤ directify₂ S [ i ∷ is ]) holds
+      ‡₂ = ⋁[ F ]-least
+            (family-of-lists S [ is ])
+            (directify₂ S [ i ∷ is ] , γ)
+
```
martinescardo, line 246: Two blank lines. One is enough.

source/Locales/Compactness/Properties.lagda
```diff
@@ -0,0 +1,716 @@
...
+
+    Ⅱ  = ap (λ - → S [ i ] ∨[ F ] -) IH
```
martinescardo, line 248: Why II before I?

martinescardo (changes requested):
Wonderful. I only have minor suggestions.

source/Locales/Compactness/Properties.lagda
```diff
@@ -0,0 +1,716 @@
...
+subset.
+
+\begin{code}
+
+ χ∙-of-Kuratowski-finite-subset-is-Kuratowski-finite
+  : (S : Fam 𝓤 ⟨ 𝒪 X ⟩)
+  → is-Kuratowski-finite (index S)
+  → is-Kuratowski-finite-subset (χ∙ S)
+ χ∙-of-Kuratowski-finite-subset-is-Kuratowski-finite S 𝕗 = ∥∥-functor † 𝕗
+  where
```
tomdjong, line 377: This should have a more conceptual (and maybe shorter?) proof: `𝕋 (χ∙ S)` should be the image of `S` and the map `index S → image S` is always a surjection, so you could simply compose the surjections, I think.

source/Locales/Compactness/Properties.lagda
```diff
@@ -0,0 +1,716 @@
...
+   † : Σ (J , h) ꞉ SubFam S , is-Kuratowski-finite J
+                            × (U ＝ ⋁[ 𝒪 X ] ⁅ U ∧[ 𝒪 X ] S [ h j ] ∣ j ∶ J ⁆)
+     → Σ (J , h) ꞉ SubFam S , is-Kuratowski-finite J
+                            × (U ≤ (⋁[ 𝒪 X ] ⁅ S [ h j ] ∣ j ∶ J ⁆)) holds
+   † (𝒥@(J , h) , 𝕗 , p) =
+    𝒥 , 𝕗 , distribute-inside-cover₁ U ⁅ S [ h j ] ∣ j ∶ J ⁆ p
+
+\end{code}
+
+Now, the forward implication, which is a bit more involved.
```
tomdjong, line 637: I think this is much simpler to prove actually: assume you have $U = \bigvee S$. By primed compactness, you get a Kuratowski finite $F \subseteq S$ with $U \le {\bigvee F}$. It remains to show that $U = {\bigvee F}$, but $\bigvee F \le \bigvee S = U$.

source/Locales/Compactness/Properties.lagda
ayberkt: There is a problem with moving this into `MLTT.List`: it uses `MLTT.Fin` which imports `MLTT.List` so it results in a cyclic dependency. I’m not sure what the best solution is. Should I create a new module?

source/Locales/Compactness/Properties.lagda
ayberkt: Fixed.

source/Locales/Compactness/Properties.lagda
martinescardo: Probably the best solution is to create MLTT.List-Properties.

ayberkt: I have now created a new module `MLTT.List-Properties` and moved these functions there.

source/Locales/Compactness/Properties.lagda
ayberkt: Fixed.

tomdjong:
Nice to see the simplifications implemented!

ayberkt:
@martinescardo This PR is now ready to be merged when the CI checks pass.

source/Locales/Compactness/Properties.lagda
ayberkt: We should also document this convention in the contribution guidelines. I was under the impression that we used capital kebab casing for module names. Do you want me add it @martinescardo? Is there anything else you would like to add?

martinescardo (approved):
I am happy with this. 

@tomdjong Do you have further comments, or can I merge this when the tests pass?

tomdjong:
Go ahead and merge it.

PR #312, opened 2024-09-04
Indentation and making more arguments implicit

tomdjong:
@IanRay11 I fixed the indentation as promised, but I also made more arguments implicit; the truncation level `n` in particular is now implicit everywhere (which caused no issues).


tomdjong:
I'll merge this into master when @IanRay11 has had a change to have a look and when the checks are completed.

source/UF/H-Levels.lagda
```diff
@@ -48,11 +50,11 @@ _is-of-hlevel_ : 𝓤 ̇ → ℕ → 𝓤 ̇
 X is-of-hlevel zero = is-contr X
 X is-of-hlevel succ n = (x x' : X) → (x ＝ x') is-of-hlevel n
 
-hlevel-relation-is-prop : {𝓤 : Universe} (n : ℕ) (X : 𝓤 ̇ )
+hlevel-relation-is-prop : {𝓤 : Universe} {n : ℕ} {X : 𝓤 ̇ }
                         → is-prop (X is-of-hlevel n)
-hlevel-relation-is-prop {𝓤} zero X = being-singleton-is-prop fe
-hlevel-relation-is-prop {𝓤} (succ n) X =
-  Π₂-is-prop fe (λ x x' → hlevel-relation-is-prop n (x ＝ x'))
+hlevel-relation-is-prop {𝓤} {zero} = being-singleton-is-prop fe
```
martinescardo, line 55: May I suggest to use "0" rather than "zero" everywhere?

source/UF/H-Levels.lagda
```diff
@@ -67,7 +69,7 @@ is-prop' : (X : 𝓤 ̇) → 𝓤  ̇
 is-prop' X = X is-of-hlevel (succ zero)
```
martinescardo, line 69: May I suggest writing `1` rather than `(succ zero)`?

source/UF/H-Levels.lagda
```diff
@@ -197,27 +193,25 @@ From univalence we can show that ℍ n is of level (n + 1), for all n : ℕ.
 
 \begin{code}
 
-equiv-preserves-hlevel : (n : ℕ) {X : 𝓤 ̇  } {Y : 𝓥 ̇  }
+equiv-preserves-hlevel : {n : ℕ} {X : 𝓤 ̇  } {Y : 𝓥 ̇  }
```
martinescardo, line 196: Please replace " ̇  }" by " ̇ }" everywhere. There is a double space. We need only one space between the upper dot and the brace. Similarly with round brackets. In the other file as well.

martinescardo (changes requested):
Thanks @IanRay11 and @tomdjong. I have few stylistic suggestions.

tomdjong:
Thanks @martinescardo! I will fix the two files as part of this PR. @IanRay11 Please follow these guidelines in your upcoming PR.

martinescardo:
Ooops. I may have created a conflict unintentionally. 

@tomdjong could you fix this for me by ignoring the changes I made in that file? Thanks!

tomdjong:
@martinescardo Should be fixed now.

martinescardo:
Thanks a lot!

source/UF/H-Levels.lagda
```diff
@@ -1,5 +1,7 @@
 Martin Escardo and Ian Ray, 06/02/2024
```
martinescardo, line 1: This authorship is kind of misleading. I don't think I had any real contribution! There still is an alternative file `HLevels` which I wrote in 2019, but that file is very short and also assumes univalence, while here you get away with function extensionality only. Moreover, there many more additions here, done all by yourself. May I suggest that you make yourself as the sole author?

(Eventually I may remove the old `HLevels` file and use yours only, when this PR is pushed.)

source/UF/Truncations.lagda
```diff
@@ -1,5 +1,7 @@
 Ian Ray, 07/23/2024
```
martinescardo, line 1: Could we please not write the date in the American system? Either write `7th September 2024` or `2024-07-07`. Because many of the contributors are based in the UK, using any convention with `99/99/99` will be ambiguous. Not in this example, because there is no month 23, of course, but still we should avoid this notation for dates. In all other files you authored as well. Thanks.

IanRay11:
@tomdjong
I'm a bit under the weather today so I will not be able to take a look at the file, but the changes sounds good to me Tom. 

@martinescardo 
As for authorship I do not care to share it since I burrowed from the existing HLevels file to get myself started. Of course it's up to you, but if you eventually plan to remove the other file I think it is fair to keep your name on this one!

source/UF/Truncations.lagda
tomdjong: I fixed this in all files.

tomdjong:
@martinescardo Shall I remove your name? We could possibly add it back when the other HLevels file gets deleted.

martinescardo:
I prefer to give full credit to @IanRay11.

martinescardo:
Great. The html rendering is now in progress. In about 10-15 it should be updated.

Many thanks, @IanRay11, for contributing this, and, @tomdjong, for the quality control.

PR #313, opened 2024-09-05
Connected Types PR

IanRay11:
This PR introduces the notion of connectedness with respect to Hlevels.

### DIFF
--- a/source/UF/ConnectedTypes.lagda
+++ b/source/UF/ConnectedTypes.lagda
@@ -1,4 +1,4 @@
-Ian Ray, 07/23/2024
+Ian Ray, 23rd July 2024
 
 We will define connected types and maps (recall our convetion for H-levels starts
 at 0). We then explore relationships, closure properties and characterizations
@@ -8,15 +8,20 @@ of interest pertaining to the concept of connectedness.
 
 {-# OPTIONS --safe --without-K #-}
 
+open import UF.FunExt
+open import UF.PropTrunc 
+
+module UF.ConnectedTypes (fe : Fun-Ext)
+                         (pt : propositional-truncations-exist)
+                          where
+                          
 open import MLTT.Spartan
 open import UF.Base
 open import UF.Embeddings
 open import UF.Equiv
 open import UF.EquivalenceExamples
 open import UF.Equiv-FunExt
-open import UF.FunExt
 open import UF.IdentitySystems
-open import UF.PropTrunc 
 open import UF.Retracts
 open import UF.Sets
 open import UF.Singleton-Properties
@@ -27,11 +32,6 @@ open import UF.Univalence
 open import UF.UA-FunExt
 open import Naturals.Addition renaming (_+_ to _+'_)
 open import Naturals.Order
-
-module UF.ConnectedTypes (fe : Fun-Ext)
-                         (pt : propositional-truncations-exist)
-                          where
-
 open import UF.ImageAndSurjection pt
 open import UF.H-Levels fe  
 open import UF.Truncations fe pt
--- a/source/UF/ConnectedTypes.lagda
+++ b/source/UF/ConnectedTypes.lagda
@@ -1,6 +1,6 @@
 Ian Ray, 23rd July 2024
 
-We will define connected types and maps (recall our convetion for H-levels starts
+We define connected types and maps (recall our convention that H-levels start
 at 0). We then explore relationships, closure properties and characterizations
 of interest pertaining to the concept of connectedness.
 
@@ -16,32 +16,25 @@ module UF.ConnectedTypes (fe : Fun-Ext)
                           where
                           
 open import MLTT.Spartan
-open import UF.Base
-open import UF.Embeddings
-open import UF.Equiv
-open import UF.EquivalenceExamples
-open import UF.Equiv-FunExt
-open import UF.IdentitySystems
-open import UF.Retracts
-open import UF.Sets
-open import UF.Singleton-Properties
-open import UF.Subsingletons
-open import UF.Subsingletons-FunExt
-open import UF.Subsingletons-Properties
-open import UF.Univalence
-open import UF.UA-FunExt
 open import Naturals.Addition renaming (_+_ to _+'_)
 open import Naturals.Order
+open import UF.Equiv
+open import UF.EquivalenceExamples
+open import UF.H-Levels fe
 open import UF.ImageAndSurjection pt
-open import UF.H-Levels fe  
+open import UF.Subsingletons
+open import UF.Subsingletons-FunExt
 open import UF.Truncations fe pt
+open import UF.Univalence
+
+open propositional-truncations-exist pt
 
 \end{code}
 
 We now define the notion of k-connectedness for types and functions with respect
 to H-levels. TODO: show that connectedness as defined elsewhere is a
 special case of k-connectedness. Connectedness typically means set connectedness.
-In terms of our Hlevel convetion it will mean 2-connectedness.
+In terms of our H-level convetion it will mean 2-connectedness.
 
 \begin{code}
 
@@ -54,54 +47,51 @@ module k-connectedness
  _is_connected : 𝓤 ̇ → ℕ → 𝓤 ̇
  X is k connected = is-contr (∥ X ∥[ k ])
 
- map_is_connected : {X : 𝓤 ̇} {Y : 𝓥 ̇} → (f : X → Y) → ℕ → 𝓤 ⊔ 𝓥 ̇
- map f is k connected = (y : codomain f) → (fiber f y) is k connected
+ _is_connected-map : {X : 𝓤 ̇} {Y : 𝓥 ̇} → (f : X → Y) → ℕ → 𝓤 ⊔ 𝓥 ̇
+ f is k connected-map = each-fiber-of f (λ - → - is k connected)
 
 \end{code}
 
-We now characterize 1-connected types as inhabited types and 1-connected maps
-as surjections.
+(y : codomain f) → (fiber f y) is k connected
+
+We characterize 1-connected types as inhabited types and 1-connected maps as
+surjections.
 
 \begin{code}
 
- open propositional-truncations-exist pt
-
- one-connected-is-inhabited : {X : 𝓤 ̇}
+ inhabited-if-one-connected : {X : 𝓤 ̇}
                             → X is 1 connected → ∥ X ∥
- one-connected-is-inhabited X-1-conn = one-trunc-to-prop-trunc (center X-1-conn)
+ inhabited-if-one-connected X-1-conn = one-trunc-to-prop-trunc (center X-1-conn)
 
- inhabited-is-one-connected : {X : 𝓤 ̇}
+ one-connected-if-inhabited : {X : 𝓤 ̇}
                             → ∥ X ∥ → X is 1 connected
- inhabited-is-one-connected x-anon =
+ one-connected-if-inhabited x-anon =
   pointed-props-are-singletons (prop-trunc-to-one-trunc x-anon) one-trunc-is-prop
 
  one-connected-iff-inhabited : {X : 𝓤 ̇}
-                             → X is 1 connected
-                             ↔ ∥ X ∥
+                             → X is 1 connected ↔ ∥ X ∥
  one-connected-iff-inhabited =
-  (one-connected-is-inhabited , inhabited-is-one-connected)
+  (inhabited-if-one-connected , one-connected-if-inhabited)
 
- one-connected-map-is-surj : {X : 𝓤 ̇} {Y : 𝓥 ̇} {f : X → Y}
-                           → map f is 1 connected
-                           → is-surjection f
- one-connected-map-is-surj f-1-con y = one-connected-is-inhabited (f-1-con y)
+ map-is-surj-if-one-connected : {X : 𝓤 ̇} {Y : 𝓥 ̇} {f : X → Y}
+                              → f is 1 connected-map → is-surjection f
+ map-is-surj-if-one-connected f-1-con y = inhabited-if-one-connected (f-1-con y)
 
- surj-is-one-connected-map : {X : 𝓤 ̇} {Y : 𝓥 ̇} {f : X → Y}
-                           → is-surjection f
-                           → map f is 1 connected
- surj-is-one-connected-map f-is-surj y = inhabited-is-one-connected (f-is-surj y)
+ map-is-one-connected-if-surj : {X : 𝓤 ̇} {Y : 𝓥 ̇} {f : X → Y}
+                              → is-surjection f → f is 1 connected-map
+ map-is-one-connected-if-surj f-is-surj y = one-connected-if-inhabited (f-is-surj y)
 
- one-connected-map-iff-surj : {X : 𝓤 ̇} {Y : 𝓥 ̇} {f : X → Y}
-                            → map f is 1 connected
-                            ↔ is-surjection f
- one-connected-map-iff-surj = (one-connected-map-is-surj , surj-is-one-connected-map)
+ map-is-one-connected-iff-surj : {X : 𝓤 ̇} {Y : 𝓥 ̇} {f : X → Y}
+                               → f is 1 connected-map ↔ is-surjection f
+ map-is-one-connected-iff-surj =
+  (map-is-surj-if-one-connected , map-is-one-connected-if-surj)
 
 \end{code}
 
-We now devlop some closure conditions pertaining to connectedness. Connectedness
+We develop some closure conditions pertaining to connectedness. Connectedness
 is closed under equivalence as expected, but more importantly connectedness
 extends below: more precisely if a type is k connected then it is l connected
-for all l ≤ k. We provide many incarnations of this fact below which may prove
+for all l ≤ k. We provide a few incarnations of this fact below which may prove
 useful. 
 
 \begin{code}
@@ -111,7 +101,7 @@ useful.
                                   → Y is k connected
                                   → X is k connected
  connectedness-closed-under-equiv e Y-con =
-   equiv-to-singleton (truncation-closed-under-equiv e) Y-con
+  equiv-to-singleton (truncation-closed-under-equiv e) Y-con
 
  contractible-types-are-connected : {X : 𝓤 ̇} {n : ℕ}
                                   → is-contr X
@@ -126,29 +116,22 @@ useful.
                                → X is (succ k) connected
                                → X is k connected
  connectedness-is-lower-closed {𝓤} {X} {k} X-succ-con =
-   equiv-to-singleton successive-truncations-equiv 
+  equiv-to-singleton successive-truncations-equiv 
                       (contractible-types-are-connected X-succ-con)
 
- connectedness-extends-to-zero : {X : 𝓤 ̇} {k : ℕ}
-                               → X is k connected
-                               → X is 0 connected
- connectedness-extends-to-zero {𝓤} {X} {0} X-con = X-con
- connectedness-extends-to-zero {𝓤} {X} {succ k} X-con =
-   connectedness-extends-to-zero (connectedness-is-lower-closed X-con)
+ connectedness-is-lower-closed-+ : {X : 𝓤 ̇} {k l : ℕ}
+                                 → X is (l +' k) connected
+                                 → X is l connected
+ connectedness-is-lower-closed-+ {𝓤} {X} {0} {l} X-con = X-con
+ connectedness-is-lower-closed-+ {𝓤} {X} {succ k} {l} X-con =
+  connectedness-is-lower-closed-+ (connectedness-is-lower-closed X-con)
 
- connectedness-step-down : {X : 𝓤 ̇} {k l : ℕ}
-                         → X is (l +' k) connected
-                         → X is l connected
- connectedness-step-down {𝓤} {X} {0} {l} X-con = X-con
- connectedness-step-down {𝓤} {X} {succ k} {l} X-con =
-   connectedness-step-down (connectedness-is-lower-closed X-con)
-
- connectedness-extends-below : {X : 𝓤 ̇} {k l : ℕ}
-                             → (l ≤ℕ k)
-                             → X is k connected
-                             → X is l connected
- connectedness-extends-below {𝓤} {X} {k} {l} o X-con =
-   connectedness-step-down (transport (λ z → X is z connected) p X-con)
+ connectedness-is-lower-closed' : {X : 𝓤 ̇} {k l : ℕ}
+                                → (l ≤ℕ k)
+                                → X is k connected
+                                → X is l connected
+ connectedness-is-lower-closed' {𝓤} {X} {k} {l} o X-con =
+  connectedness-is-lower-closed-+ (transport (λ z → X is z connected) p X-con)
   where
    m : ℕ
    m = pr₁ (subtraction l k o)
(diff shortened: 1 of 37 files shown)


IanRay11:
@tomdjong note that in connectedness-extends-below I use a projection since afaik there isn't a named projection for it in the order file. Also we may want a lemma to simplify the proof of inhabited-id-conn-to-conn.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
+Ian Ray, 23rd July 2024
+
+We will define connected types and maps (recall our convetion for H-levels starts
```
tomdjong, line 3: * _convention_ (Applies elsewhere too. Maybe enable a spell-checker to catch these and other misspellings.)
* _that H-levels_
* _We define_ sounds more natural to me and is in line with _we explore_.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+{-# OPTIONS --safe --without-K #-}
+
+open import UF.FunExt
+open import UF.PropTrunc 
+
+module UF.ConnectedTypes (fe : Fun-Ext)
+                         (pt : propositional-truncations-exist)
+                          where
+                          
+open import MLTT.Spartan
```
tomdjong, line 18: Sort and remove unused imports.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+
+We now define the notion of k-connectedness for types and functions with respect
+to H-levels. TODO: show that connectedness as defined elsewhere is a
+special case of k-connectedness. Connectedness typically means set connectedness.
+In terms of our Hlevel convetion it will mean 2-connectedness.
+
+\begin{code}
+
+module k-connectedness
+        (te : H-level-truncations-exist)
```
tomdjong, line 49: This applies to everything you do, so I suggest making it a top-level parameter, i.e. an assumption for `UF.ConnectedTypes` like `fe` and `pt`.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+       where
+
+ open H-level-truncations-exist te
+
+ _is_connected : 𝓤 ̇ → ℕ → 𝓤 ̇
+ X is k connected = is-contr (∥ X ∥[ k ])
+
+ map_is_connected : {X : 𝓤 ̇} {Y : 𝓥 ̇} → (f : X → Y) → ℕ → 𝓤 ⊔ 𝓥 ̇
+ map f is k connected = (y : codomain f) → (fiber f y) is k connected
```
tomdjong, line 58: Please use `each-fiber-of` instead.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+
+\end{code}
+
+We now characterize 1-connected types as inhabited types and 1-connected maps
+as surjections.
+
+\begin{code}
+
+ open propositional-truncations-exist pt
```
tomdjong, line 67: Do this at the top after the imports somewhere.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+
+ one-connected-is-inhabited : {X : 𝓤 ̇}
```
tomdjong, line 69: Clearer: `inhabited-if-one-connected`. Please follow this convention everywhere.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+ one-connected-is-inhabited X-1-conn = one-trunc-to-prop-trunc (center X-1-conn)
+
+ inhabited-is-one-connected : {X : 𝓤 ̇}
+                            → ∥ X ∥ → X is 1 connected
+ inhabited-is-one-connected x-anon =
+  pointed-props-are-singletons (prop-trunc-to-one-trunc x-anon) one-trunc-is-prop
+
+ one-connected-iff-inhabited : {X : 𝓤 ̇}
+                             → X is 1 connected
+                             ↔ ∥ X ∥
```
tomdjong, line 80: I would merge lines 79 and 80.

tomdjong: Same with 95 and 96.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+ surj-is-one-connected-map f-is-surj y = inhabited-is-one-connected (f-is-surj y)
+
+ one-connected-map-iff-surj : {X : 𝓤 ̇} {Y : 𝓥 ̇} {f : X → Y}
+                            → map f is 1 connected
+                            ↔ is-surjection f
+ one-connected-map-iff-surj = (one-connected-map-is-surj , surj-is-one-connected-map)
+
+\end{code}
+
+We now devlop some closure conditions pertaining to connectedness. Connectedness
```
tomdjong, line 101: _develop_

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+                (λ x → ap ∣_∣[ n ] (C x))
+
+ connectedness-is-lower-closed : {X : 𝓤 ̇} {k : ℕ}
+                               → X is (succ k) connected
+                               → X is k connected
+ connectedness-is-lower-closed {𝓤} {X} {k} X-succ-con =
+   equiv-to-singleton successive-truncations-equiv 
+                      (contractible-types-are-connected X-succ-con)
+
+ connectedness-extends-to-zero : {X : 𝓤 ̇} {k : ℕ}
```
tomdjong, line 132: This lemma is a little strange because _any_ type is `0`-connected, essentially by definition.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+                               → X is k connected
+                               → X is 0 connected
+ connectedness-extends-to-zero {𝓤} {X} {0} X-con = X-con
+ connectedness-extends-to-zero {𝓤} {X} {succ k} X-con =
+   connectedness-extends-to-zero (connectedness-is-lower-closed X-con)
+
+ connectedness-step-down : {X : 𝓤 ̇} {k l : ℕ}
```
tomdjong, line 139: I suggest calling this `connectedness-is-lower-closed-+`.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+                         → X is (l +' k) connected
+                         → X is l connected
+ connectedness-step-down {𝓤} {X} {0} {l} X-con = X-con
+ connectedness-step-down {𝓤} {X} {succ k} {l} X-con =
+   connectedness-step-down (connectedness-is-lower-closed X-con)
+
+ connectedness-extends-below : {X : 𝓤 ̇} {k l : ℕ}
```
tomdjong, line 146: I suggest calling this `connectedness-is-lower-closed'`. The phrase "extends below" is a bit mysterious in my opinion.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+       l +' m   ∎
+
+\end{code}
+
+We can characterize connected types in terms of inhabitedness and connectedness
+of the identity type at the level below. We will assume univalence when neccesary.
+
+\begin{code}
+
+ conn-to-inhabited : {X : 𝓤 ̇} {n : ℕ}
```
tomdjong, line 166: I would call this `inhabited-if-connected` and there is a more conceptual proof using `inhabited-if-one-connected`.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+                   → X is (succ n) connected
+                   → ∥ X ∥
+ conn-to-inhabited {_} {_} {0} X-conn =
+  center (equiv-to-singleton' one-trunc-≃-prop-trunc X-conn)
+ conn-to-inhabited {_} {_} {succ n} X-conn =
+  conn-to-inhabited (connectedness-is-lower-closed X-conn)
+
+ conn-to-id-conn : {X : 𝓤 ̇} {n : ℕ}
```
tomdjong, line 174: I would rename this to `connected-types-are-locally-connected`.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+                 → (ua : is-univalent 𝓤)
+                 → X is (succ n) connected
+                 → (x y : X) → (x ＝ y) is n connected
+ conn-to-id-conn {_} {_} {n} ua X-conn x y =
+  equiv-to-singleton (eliminated-trunc-identity-char ua)
+                     (is-prop-implies-is-prop' (singletons-are-props X-conn)
+                                               ∣ x ∣[ succ n ]
+                                               ∣ y ∣[ succ n ])
+
+ conn-to-inhabited-id-conn : {X : 𝓤 ̇} {n : ℕ}
```
tomdjong, line 184: This then becomes `connected-types-are-inhabited-and-locally-connected`.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+                           → (ua : is-univalent 𝓤)
+                           → X is (succ n) connected
+                           → ∥ X ∥ × ((x y : X) → (x ＝ y) is n connected)
+ conn-to-inhabited-id-conn ua X-conn =
+  (conn-to-inhabited X-conn , conn-to-id-conn ua X-conn)
+
+ inhabited-id-conn-to-conn : {X : 𝓤 ̇} {n : ℕ}
```
tomdjong, line 191: `inhabited-and-locally-connected-types-are-connected`.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+         (λ x → (∣ x ∣[ succ (succ n) ]
+          , ∥∥ₙ-ind (λ v → hlevels-are-upper-closed
+                            (λ p q → ∥∥ₙ-hlevel ∣ x ∣[ succ (succ n) ] v p q)) 
+                    (λ y → forth-trunc-id-char ua (center (id-conn x y)))))
+         anon-x
+
+ connected-characterization : {X : 𝓤 ̇} {n : ℕ}
+                            → (ua : is-univalent 𝓤)
+                            → X is (succ n) connected
+                            ↔ ∥ X ∥ × ((x y : X) → (x ＝ y) is n connected)
```
tomdjong, line 208: Might be nice to introduce the definition `_is-locally_connected`.

source/UF/ConnectedTypes.lagda
tomdjong: Suggestion: maybe `_is_connected-map` is nicer than `map_is_connected`?

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+ conn-to-inhabited : {X : 𝓤 ̇} {n : ℕ}
+                   → X is (succ n) connected
+                   → ∥ X ∥
+ conn-to-inhabited {_} {_} {0} X-conn =
+  center (equiv-to-singleton' one-trunc-≃-prop-trunc X-conn)
+ conn-to-inhabited {_} {_} {succ n} X-conn =
+  conn-to-inhabited (connectedness-is-lower-closed X-conn)
+
+ conn-to-id-conn : {X : 𝓤 ̇} {n : ℕ}
+                 → (ua : is-univalent 𝓤)
```
tomdjong, line 175: You don't need to name the univalence assumption.

tomdjong: This applies elsewhere too.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+         anon-x
+
+ connected-characterization : {X : 𝓤 ̇} {n : ℕ}
+                            → (ua : is-univalent 𝓤)
+                            → X is (succ n) connected
+                            ↔ ∥ X ∥ × ((x y : X) → (x ＝ y) is n connected)
+ connected-characterization ua =
+  (conn-to-inhabited-id-conn ua , inhabited-id-conn-to-conn ua)
+
+ ap-is-less-connected : {X : 𝓤 ̇} {Y : 𝓥 ̇} {n : ℕ} {x x' : X}
```
tomdjong, line 212: It's nicer in practice and conceptually cleaner to introduce `{x x' : X}` after introducing `f` and the fact that it's `n + 1`-connected. (That way you can forget about all the other implicit variables.)

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,223 @@
...
+                      → (ua : is-univalent (𝓤 ⊔ 𝓥))
+                      → (f : X → Y)
+                      → map f is (succ n) connected
+                      → map (ap f {x} {x'}) is n connected
+ ap-is-less-connected {_} {_} {_} {_} {n} {x} {x'} ua f f-conn p =
+  equiv-to-singleton (truncation-closed-under-equiv (fiber-of-ap-≃ f p))
+                     (conn-to-id-conn ua (f-conn (f x')) (x , p) (x' , refl))
+
+\end{code}
+
```
tomdjong, line 222: Remove these blank lines at the end of the file.

tomdjong (changes requested):
Thanks @IanRay11!

source/UF/ConnectedTypes.lagda
IanRay11: I need H-Levels (which depends on Fun-Ext) imported to use this right? I considered this but didn't come up with a solution that avoided having to stack modules

source/UF/ConnectedTypes.lagda
IanRay11: I'll remove it.

IanRay11: I think I thought it was a base case for something, but it didn't end up being needed.

source/UF/ConnectedTypes.lagda
IanRay11: Okay I'll check it out

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,212 @@
...
+of interest pertaining to the concept of connectedness.
+
+\begin{code}
+
+{-# OPTIONS --safe --without-K #-}
+
+open import UF.FunExt
+open import UF.PropTrunc 
+
+module UF.ConnectedTypes (fe : Fun-Ext)
```
tomdjong, line 14: Can you please indent like this instead?
```
module UF.ConnectedTypes 
        (fe : Fun-Ext)
        (pt : propositional-truncations-exist)
       where
```
Please do this in your other files too.

source/UF/ConnectedTypes.lagda
tomdjong: Ah I see. Let's leave it like this for now then.

tomdjong:
@IanRay11 Sorry for commenting on the other file, but it only just caught my eye: In `UF.Truncations.lagda`, the existence of propositional truncations should not be a global assumption, but rather a local one (when comparing truncations). But the general truncation shouldn't depend on prop. truncations existing.

IanRay11:
Do you want me to open another PR fixing this? @tomdjong

tomdjong:
@IanRay11 You can just do it in this PR.

source/UF/ConnectedTypes.lagda
IanRay11: For all modules? What if there is only one module param?

tomdjong: Ah right, this should only need Fun-Ext now. You can also use a single line if there's only one assumption.

tomdjong: The `where` should be on that same line.

IanRay11: Well I do use pt in this file a fair bit because of all the inhabitedness. So I plan to still use it globally here. I was gonna fix it in my other files if necesary.

tomdjong: But you can just use the 1-truncation? I'm not sure which is more natural.

IanRay11: I mean I have shown they are equivalent but I think it's more clear to use the prop trunc. Would it be worthwhile to show the truncations-exist omplies prop-trunc-exists?

IanRay11:
> @IanRay11 Sorry for commenting on the other file, but it only just caught my eye: In `UF.Truncations.lagda`, the existence of propositional truncations should not be a global assumption, but rather a local one (when comparing truncations). But the general truncation shouldn't depend on prop. truncations existing.

IanRay11:
@tomdjong I did not mean to close

IanRay11:
I made prop-trunc-exists a local assumption in the Truncations file. For now I have left it a global assumption in ConnectedTypes but am open to continue that discussion when you review tomorrow.

IanRay11:
That is enlightening. In this case I propose we leave it pt as a global assumption. 

Apologies I forgot tomorrow was the start of the weekend.

tomdjong:
> In this case I propose we leave it pt as a global assumption.

Why? If you're going to assume general truncations, it's conceptually strange to separately assume prop. truncations.

IanRay11:
It feels conceptually clearer to me to use prop trunc notation if we are using the word 'inhabited' in the name of the theorems that follow.

IanRay11:
Also if we use 1-truncation instead the proofs would need to reworked. This isn't a problem from the standpoint of effort but reworking the proofs would likely require me to call in prop-trunc anyway since it is required to show the two are equivalent.

tomdjong:
Sorry, there is a miscommunication here. I agree it's good to use the propositional truncation, but its *existence* should be derived from the existence of general propositional truncations, i.e. you *prove* `propositional-truncations-exist`.

Does that make sense?

IanRay11:
> Conceptually, it's definitely the right thing to do: we also get function extensionality from univalence in many places, rather than assuming it separately. 
> 
> 
> 
> However, this can sometimes lead to problems when Agda is expecting an "abstract" `pt` and not a particular construction of it. (This happened with set replacement and the existence of set quotients in the past.) But I suggest to try to the conceptually right thing first to see if you get such issues.

I managed to not read the last sentence 😅

I'll give it a shot.

IanRay11:
In truncations I couldn't replace the local assumption of pt below my recent addition becuase I am inside the H-level-truncations-exist record and don't actually have a term te to work with. Do you have any suggestions? @tomdjong

source/UF/Truncations.lagda
```diff
@@ -150,6 +150,23 @@ groupoid).
...
 
+We show that the existence of propositional truncation follows from the existence
+of general truncations. Notice this implication manifests as a function between
+record types.
+
+\begin{code}
+
+ prop-trunc-follows-from-gen-trunc : H-level-truncations-exist
+                                   → propositional-truncations-exist
+ prop-trunc-follows-from-gen-trunc te = record
```
tomdjong, line 161: You're not using the assumption `te` here. Rather you're within the scope of the original `H-level-truncations-exist` record.
I would also rename this to `H-level-truncations-give-propositional-truncations`.

tomdjong: You can simply exit the scope by indenting one level less.

IanRay11: Oh right. So should I do this outside of the record?

tomdjong: You can also do it within the record, but then its type is `propositional-truncations-exist`, so I think doing it outside the record is clearer.

tomdjong: Within the record is like going outside and then doing `open H-level-truncations-exist te`.

IanRay11: > You can simply exit the scope by indenting one level less.

When I indent it gives me an error on the code that follows that is supposed to be back within the scope of the record.

tomdjong: Like I said, you have to open the record with your assumption.

tomdjong: It works just as with modules.

IanRay11: I understand that I exited the scope then opened the record with my assumption using a where statement. The problem is that indenting below doesn't put me back in the scope right? It would be nice to prove this and then use it in the file to avoid an pt assumption at all.

tomdjong: I don't understand `avoid any pt assumption at all`? But if you want the existence of propositional truncations to become available as soon as you have general truncations, then you can prove this fact within the scope of general truncations record, but you just simplify change its type to `propositional-truncations-exist`.

Does that help?

IanRay11: Ahh I guess I see your point. Let me push what I decided to go with and see what you think.

tomdjong:
@IanRay11 Yes, this is what I had in mind, although `pt = H-level-truncations-give-propositional-truncations te` should be `private`, juts like we do when we get `fe : Fun-Ext` from `Univalence` in a file that assumes it.

IanRay11:
Right

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,210 @@
...
+        (fe : Fun-Ext)
+       where
+                          
+open import MLTT.Spartan
+open import Naturals.Addition renaming (_+_ to _+'_)
+open import Naturals.Order
+open import UF.Equiv
+open import UF.EquivalenceExamples
+open import UF.H-Levels fe
+
```
tomdjong, line 23: Please remove this blank line.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,210 @@
...
+open import UF.Subsingletons-FunExt
+open import UF.Truncations fe
+open import UF.Univalence
+
+\end{code}
+
+We now define the notion of k-connectedness for types and functions with respect
+to H-levels. TODO: show that connectedness as defined elsewhere is a
+special case of k-connectedness. Connectedness typically means set connectedness.
+In terms of our H-level convetion it will mean 2-connectedness.
```
tomdjong, line 35: _convention_ (Please check this everywhere!)

source/UF/ConnectedTypes.lagda
```diff
@@ -30,9 +29,8 @@ open import UF.Univalence
 \end{code}
 
 We now define the notion of k-connectedness for types and functions with respect
-to H-levels. TODO: show that connectedness as defined elsewhere is a
-special case of k-connectedness. Connectedness typically means set connectedness.
-In terms of our H-level convetion it will mean 2-connectedness.
+to H-levels. TODO: show that connectedness as defined elsewhere in the library is
+a special case of k-connectedness. Connectedness typically means set connectedness, by our convention it will mean 2-connectedness.
```
tomdjong, line 33: Don't forget to wrap lines at 80 characters. (Sorry...)

tomdjong:
@martinescardo Do you wish to have a look at this yourself?

martinescardo:
Reading the code, which is otherwise very nice, is very confusing as the indices are shifted by 2.

I wonder if this problem can be created using https://agda.readthedocs.io/en/latest/language/literal-overloading.html#literal-overloading

The idea would be to create a new type ℕ₋₂, in a record with exactly one field, of type ℕ, and then use a `from` function as explained in the above link.

source/UF/ConnectedTypes.lagda
```diff
@@ -0,0 +1,209 @@
...
+
+ connectedness-is-lower-closed-+ : {X : 𝓤 ̇} {k l : ℕ}
+                                 → X is (l +' k) connected
+                                 → X is l connected
+ connectedness-is-lower-closed-+ {𝓤} {X} {0} {l} X-con = X-con
+ connectedness-is-lower-closed-+ {𝓤} {X} {succ k} {l} X-con =
+  connectedness-is-lower-closed-+ (connectedness-is-lower-closed X-con)
+
+ connectedness-is-lower-closed' : {X : 𝓤 ̇} {k l : ℕ}
+                                → (l ≤ℕ k)
```
martinescardo, line 129: You should be able to write `l ≤ k` if you import `Notation.Order`.

martinescardo (approved):
Nice! You can address my comment about `l ≤ℕ k` in a future PR.

martinescardo:
@tomdjong You can merge this if you are happy with it.

source/UF/ConnectedTypes.lagda
martinescardo: Also you don't need the brackets.

PR #315, opened 2024-09-06
Unused imports

tomdjong:
@martinescardo I will continue cleaning in this PR. Please leave it open for now.

### DIFF
--- a/source/TWA/BanachFixedPointTheorem.lagda
+++ b/source/TWA/BanachFixedPointTheorem.lagda
@@ -12,10 +12,8 @@ module TWA.BanachFixedPointTheorem (fe : FunExt) where
 
 open import MLTT.Spartan
 open import CoNaturals.Type hiding (min)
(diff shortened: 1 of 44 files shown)


ayberkt:
It might also be a good idea to include this script in the CI checks. We can avoid the addition of new modules with unused imports this way.

tomdjong:
@ayberkt We should only run it on new (or maybe modified) files then because this has the potential of being very slow.

martinescardo:
> @ayberkt We should only run it on new (or maybe modified) files then because this has the potential of being very slow.

Agreed!

ayberkt:
> We should only run it on new (or maybe modified) files then

That is a good point. Running it only on modified files should be possible.

How does the script detect unused imports?

tomdjong:
By uncommenting line by line and seeing if it compiles. The script file should be readable and essentially self-explanatory.

PR #320, opened 2024-09-11
Properties of ordinal multiplication and reverse lexicographic order for α ×ₒ β

tomdjong:
Notably multiplication by an ordinal strictly greater than 0 is left-cancellable.
We also split `ArithmeticProperties.lagda` into `AdditionProperties.lagda` and `MultiplicationProperties.lagda`.

cc @fredrikNordvallForsberg


source/Ordinals/MultiplicationProperties.lagda
```diff
@@ -0,0 +1,489 @@
...
+         eq = f  (a' , b)             ＝⟨ IH (a' , b) (inr (refl , q)) ⟩
+              f' (a' , b)             ＝⟨ refl ⟩
+              (a' , pr₂ (f (a₀ , b))) ＝⟨ ap (a' ,_) (r ⁻¹) ⟩
+              (a' , c')               ∎
+
+\end{code}
+
+The following result states that multiplication for ordinals can be
+cancelled on the left. Interestingly, Andrew Swan [Swa18] proved that
+the corresponding result for mere sets is not provable constructively
```
martinescardo: I am not requesting a change. But "mere sets" can be confusing given that some people use "merely" to mean propositional truncation. You could just say "sets".

source/Ordinals/MultiplicationProperties.lagda
```diff
@@ -0,0 +1,489 @@
...
+   II β IH γ e = Extensionality (OO 𝓤) β γ (to-≼ III) (to-≼ IV)
+    where
+     III : (b : ⟨ β ⟩) → (β ↓ b) ⊲ γ
+     III b = let (c , eq) = I β γ  e     b in (c , IH b (γ ↓ c) eq)
+     IV  : (c : ⟨ γ ⟩) → (γ ↓ c) ⊲ β
+     IV  c = let (b , eq) = I γ β (e ⁻¹) c in (b , (IH b (γ ↓ c) (eq ⁻¹) ⁻¹))
+
+\end{code}
+
+Finally, multiplication satisfies the expected recursive equations (which
```
martinescardo, line 429: I am not requesting any change. I am just wondering. You show that both addition and multiplication satisfy the usual classical equations, which is very nice. Although they don't *define* addition and multiplication in this setting, we may wonder whether these equations have addition and multiplication as unique solutions. I wonder if you've thought about this.

martinescardo (approved):
Very nice!

source/Ordinals/MultiplicationProperties.lagda
martinescardo: Yeah, I think that, for now, it would interesting to add this as comment, because it is a natural question that I am sure many other people will ask themselves.

tomdjong: We live in the future now: https://github.com/martinescardo/TypeTopology/pull/322 :)

PR #326, opened 2024-09-20
Cache generated agdai files between CI runs

ayberkt:


ayberkt:
@martinescardo I have now updated the GitHub action to save the generated `.agdai` files between builds.

I tested this with commit [46df764](https://github.com/martinescardo/TypeTopology/pull/326/commits/46df76437eb64a001d3b66a9309328fb0936a8fe), which passed the checks in only two minutes!

PR #327, opened 2024-10-08
Add more multiplication properties

TheoXIII:
This adds more properties regarding multiplication to Naturals/Order.lagda:
- Proof that if m <= n and x <= y, then m * x < = n  * y.
- Proof that if m < n and x < y, then m  *  x  <  n  *  y.


TheoXIII:
Closed for now as I wish to include this proof in my final year project.

PR #332, opened 2025-01-02
Script for generating Makefile

ayberkt:
@martinescardo As promised a while ago, I added a script to generate a Makefile from the dependency graph.

This PR adds:
  1. a small Haskell program, namely `admin-utilities/GenerateMakefile.hs`, which expects the dependency graph to be in the file `admin-utilities/dependency_graph.dot` and translates it into a Makefile.
  2. a script `admin-utilities/generate-makefile.sh`, which first writes the dependency graph into `admin-utilities/dependency_graph.dot` and then runs `GenerateMakefile.hs`.

So to generate the Makefile, one can simply run `./admin-utilities/generate-makefile.sh` from the project root directory.


martinescardo:
Thanks!

PR #334, opened 2025-01-15
Adding a proof of the universal property of pullbacks that avoids fun…

IanRay11:
…ction extensionality

### DIFF
--- a/source/Slice/Construction.lagda
+++ b/source/Slice/Construction.lagda
@@ -87,6 +87,27 @@ to-span {𝓤} {𝓥} {𝓦} {𝓤'} {A} {B} {C} f g X =
        → (Σ j ꞉ (X → A × B) , f ∘ pr₁ ∘ j ∼ g ∘ pr₂ ∘ j)
      ψ (k , l , H) = ((λ x → (k x , l x)) , H)
 
+\end{code}
+
+We include an alternate proof that does not use function extensionality.
+
+\begin{code}
+
+→-pullback-≃' : {A : 𝓤 ̇ } {B : 𝓥 ̇ } {C : 𝓦 ̇ }
+                (f : A → C) (g : B → C)
+                (X : 𝓤' ̇ )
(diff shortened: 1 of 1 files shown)


source/Slice/Construction.lagda
```diff
@@ -87,6 +87,27 @@ to-span {𝓤} {𝓥} {𝓦} {𝓤'} {A} {B} {C} f g X =
        → (Σ j ꞉ (X → A × B) , f ∘ pr₁ ∘ j ∼ g ∘ pr₂ ∘ j)
      ψ (k , l , H) = ((λ x → (k x , l x)) , H)
 
+\end{code}
+
```
martinescardo, line 91: Add "Added by Ian Ray 15th Jan 2025.".

source/Slice/Construction.lagda
```diff
@@ -87,6 +87,27 @@ to-span {𝓤} {𝓥} {𝓦} {𝓤'} {A} {B} {C} f g X =
...
+ where
+  ϕ : (X → pullback f g) → to-span f g X
+  ϕ u = ppr₁ ∘ u , ppr₂ ∘ u , ppr₃ ∘ u
+  ψ : to-span f g X → (X → pullback f g)
+  ψ (h , k , H) x = h x , k x , H x
+  ϕ-ψ : ϕ ∘ ψ ∼ id
+  ϕ-ψ (h , k , H) = refl
+  ψ-ϕ : ψ ∘ ϕ ∼ id
+  ψ-ϕ u = refl
+
```
martinescardo, line 110: Add "End of addition.".

martinescardo (changes requested):
Thanks a lot. It is common practice to locally add authors and modification dates in TypeTopology.

(For example, if we rename a file, git loses the record of modifications and authors. Also, it is much easier to see in the file, rather than by querying git, who did what then. And authors deserve credit.)

IanRay11:
Not sure why the checks failed here. I'll check back later.

PR #341, opened 2025-02-14
Create the index for the internal effectful forcing paper

ayberkt:
Although the main purpose of this PR is to create an index for the internal effectful forcing paper, it also implements some additional changes such as simplifications to some of the proofs and organisational improvements.


martinescardo:
There are conflicts, @ayberkt

PR #343, opened 2025-03-18
Index for our revised domain theory paper

tomdjong:
cc @martinescardo


martinescardo (approved):
Thanks. Perhaps it would be nice to include the sections and subsections titles, for people who look at this without (first) looking at the paper. I'll leave it for you to decide. Otherwise it is very faithful to the paper.

PR #346, opened 2025-03-26
The double negation image

fredrik-bakke:
Adds the double negation image, as well as some basic properties.

### DIFF
--- a/source/NotionsOfDecidability/Decidable.lagda
+++ b/source/NotionsOfDecidability/Decidable.lagda
@@ -417,14 +417,16 @@ is-decidable-map f = each-fiber-of f is-decidable
                 → is-decidable-map g
                 → is-decidable-map f
                 → is-decidable-map (g ∘ f)
-∘-decidable-map f g H G F x =
- cases
-  (λ u →
+∘-decidable-map f g H G F x = cases positive-case negative-case (G x)
+ where
+  positive-case : (u : fiber g x) → is-decidable (fiber (g ∘ f) x)
+  positive-case u =
    decidable-↔
     ((λ v → (v .pr₁) , (ap g (v .pr₂) ∙ u .pr₂)) ,
-     (λ w → w .pr₁ , H (w .pr₂ ∙ (u .pr₂)⁻¹)))
-    (F (u .pr₁)))
-  (λ α → inr (λ t → α (f (t .pr₁) , t .pr₂)))
-  (G x)
+     (λ w → (w .pr₁) , H (w .pr₂ ∙ (u .pr₂)⁻¹)))
+    (F (u .pr₁))
+
+  negative-case : (nu : ¬ (fiber g x)) → is-decidable (fiber (g ∘ f) x)
+  negative-case α = inr (λ t → α (f (t .pr₁) , t .pr₂))
 
 \end{code}
--- a/source/NotionsOfDecidability/Decidable.lagda
+++ b/source/NotionsOfDecidability/Decidable.lagda
@@ -296,8 +296,7 @@ all-types-are-¬¬-decidable X h = claim₂ claim₁
   claim₂ nx = h (inr nx)
 
 ¬¬-stable-if-decidable : (X : 𝓤 ̇ ) → is-decidable X → ¬¬-stable X
-¬¬-stable-if-decidable X (inl  x) = λ _ → x
-¬¬-stable-if-decidable X (inr nx) = λ h → 𝟘-elim (h nx)
+¬¬-stable-if-decidable X = ¬¬-elim
 
 \end{code}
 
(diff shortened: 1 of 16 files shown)


source/NotionsOfDecidability/Decidable.lagda
```diff
@@ -296,8 +296,7 @@ all-types-are-¬¬-decidable X h = claim₂ claim₁
   claim₂ nx = h (inr nx)
 
 ¬¬-stable-if-decidable : (X : 𝓤 ̇ ) → is-decidable X → ¬¬-stable X
-¬¬-stable-if-decidable X (inl  x) = λ _ → x
-¬¬-stable-if-decidable X (inr nx) = λ h → 𝟘-elim (h nx)
+¬¬-stable-if-decidable X = ¬¬-elim
```
martinescardo, line 299: Good.

source/NotionsOfDecidability/Decidable.lagda
```diff
@@ -403,3 +402,31 @@ module propositional-truncation-of-decidable-type
...
+
+∘-decidable-map : {X : 𝓤 ̇ } {Y : 𝓥 ̇ } {Z : 𝓦 ̇ } (f : X → Y) (g : Y → Z)
+                → left-cancellable g
+                → is-decidable-map g
+                → is-decidable-map f
+                → is-decidable-map (g ∘ f)
+∘-decidable-map f g H G F x = cases positive-case negative-case (G x)
+ where
+  positive-case : (u : fiber g x) → is-decidable (fiber (g ∘ f) x)
+  positive-case u =
```
martinescardo, line 423: Avoid projections with the use of pattern matching, to try to make the code more readable. There are more places below where this should also be done.

source/TypeTopology/Density.lagda
```diff
@@ -97,3 +100,138 @@ module _ {𝓤 𝓥} {X : 𝓤 ̇ } {Y : 𝓥 ̇ } where
...
+
+¬¬-image-factorization : {X : 𝓤 ̇ } {Y : 𝓥 ̇ } (f : X → Y)
+                       → f ∼ ¬¬-restriction f ∘ ¬¬-corestriction f
+¬¬-image-factorization f x = refl
+
+¬¬-corestrictions-are-dense : ({A : 𝓤 ⊔ 𝓥 ̇ } → is-prop (¬ A))
+                            → {X : 𝓤 ̇ } {Y : 𝓥 ̇ } (f : X → Y)
+                            → is-dense (¬¬-corestriction f)
+¬¬-corestrictions-are-dense negations-are-props f ((y , nnq) , np) =
+  nnq λ p → np (p .pr₁ , to-Σ-＝ (p .pr₂ , negations-are-props _ nnq))
```
martinescardo, line 143: Use pattern matching on lambda.

source/TypeTopology/Density.lagda
```diff
@@ -97,3 +100,138 @@ module _ {𝓤 𝓥} {X : 𝓤 ̇ } {Y : 𝓥 ̇ } where
...
+
+¬¬-restrictions-are-embeddings : ({A : 𝓤 ⊔ 𝓥 ̇ } → is-prop (¬ A))
+                               → {X : 𝓤 ̇ } {Y : 𝓥 ̇ } (f : X → Y)
+                               → is-embedding (¬¬-restriction f)
+¬¬-restrictions-are-embeddings negations-are-props f = pr₁-is-embedding
+                                                        (λ y →
+                                                         negations-are-props)
+
+¬¬-restrictions-are-left-cancellable : ({A : 𝓤 ⊔ 𝓥 ̇ } → is-prop (¬ A))
```
martinescardo, line 152: Because you are using `({A : 𝓤 ⊔ 𝓥 ̇ } → is-prop (¬ A))` multiple times, I propose you give a name to this, for example `negations-are-props`. Maybe place it in the FunExt file, and include a proof that FunExt implies that.

source/TypeTopology/Density.lagda
```diff
@@ -97,3 +100,138 @@ module _ {𝓤 𝓥} {X : 𝓤 ̇ } {Y : 𝓥 ̇ } where
...
+¬¬-restrictions-are-left-cancellable negations-are-props f =
+ embeddings-are-lc (¬¬-restriction f)
+  (¬¬-restrictions-are-embeddings negations-are-props f)
+
+¬¬-restrictions-are-¬¬-stable : ({A : 𝓤 ⊔ 𝓥 ̇ } → is-prop (¬ A))
+                              → {X : 𝓤 ̇ } {Y : 𝓥 ̇ } (f : X → Y)
+                              → is-¬¬-stable-map (¬¬-restriction f)
+¬¬-restrictions-are-¬¬-stable negations-are-props f y nnip = ((y , a) , refl)
+ where
+ a : y ∈¬¬-image f
```
martinescardo, line 164: Indentation.

martinescardo (changes requested):
Thanks for this contribution. Nice!

source/TypeTopology/Density.lagda
fredrik-bakke: The proof that funext implies that negations are propositions already has the name `negations-are-props` in the cited file, so the axiom should probably have a different name.
I also have a different question, I see few places where funext is assumed but only `negations-are-props` is used. Should I refactor those places to assume only `negations-are-props` also?

martinescardo: Just leave the uses of funext. I don't think it is a worthwhile generalization, as I don't know any natural model where negations are props but funext fails. Also, have you thought how much function extensionality one can recover from negations being props?

fredrik-bakke: It was suggested to me that Gödel's dialectica may be a natural place where negations are propositions, but function extensionality fails. I don't know much about this topic beyond that however.

fredrik-bakke: I factored the results on "negations are propositions" to the bottom of `Subsingletons-FunExt`. However, note that some of the results were not about funext but propext instead.

Again, apologies if this is not proper etiquette.

fredrik-bakke:
I moved all of the current lemmas to `TypeTopology.DenseMapsProperties` as you asked. I also renamed "decidable map" to "complemented map", which seems more in line with the library's terminology.

fredrik-bakke:
Let me know if further changes are needed, @martinescardo. I'm still thinking to leave Lavwere's fixed point theorem for a second pull request.

martinescardo (approved):
Thanks!

PR #347, opened 2025-03-27
Uniformly decidable type families

fredrik-bakke:


source/TypeTopology/CompactTypes.lagda
```diff
@@ -1233,5 +1233,22 @@ Compact-gives-Compact' C A _ = C A
...
 
+Added 27th March 2025 by Fredrik Bakke. A decidable type with double negation
+dense equality is compact. In particular, connected types are compact.
+
+\begin{code}
+
+decidable-types-with-double-negation-dense-equality-are-Compact : {X : 𝓤 ̇ }
+                                                                → is-decidable X
+                                                                → ((a b : X)
+                                                                 → ¬¬ (a ＝ b))
```
martinescardo, line 1244: What would be a type which is not a proposition but that any two of its element are ¬¬-equal?

fredrik-bakke: Any connected type

fredrik-bakke: I see, thank you for that explanation!

fredrik-bakke: By the same argument, every boolean-valued map on a type with double negation dense equality is constant, I believe. At least, the formalization goes along those lines.

fredrik-bakke: Yes, precisely by `totally-separated-types-are-¬¬-separated` this is true.

fredrik-bakke: Then I consider this PR redundant. Thank you again for the explanation.

fredrik-bakke: Have you considered the relation between the totally separated reflection and other notions of compactness, such as "double negation stability compactness"? I have an analogous proof considering those for types with double negation dense equality.

PR #350, opened 2025-04-09
Alternative pushout presentation

IanRay11:
I want to record this alternative presentation somewhere as a draft. There are two big problems.

1) The file does not type check fully. It hangs at line 275 for an unknown reason and lossy unification isn't helping.
2) There are still unfilled holes.

One benefit of the approach taken in this file is that it closes the loop on the equivalence between

- Dependent universal property of the pushout
- Induction principle and computation rules
- Recursion principle and computation rules together with uniqueness 
- Universal property of the pushout

I may revisit this in the future.

### DIFF
--- a/source/UF/Pushouts2.lagda
+++ b/source/UF/Pushouts2.lagda
@@ -1,0 +1,757 @@
+\begin{code}
+
+{-# OPTIONS --safe --without-K --lossy-unification #-}
+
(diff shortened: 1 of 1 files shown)


martinescardo:
I think it is safe to close this attempt. It will not be deleted, so that it will still be available for future reference.

PR #351, opened 2025-04-15
Cantor–Schröder–Bernstein follows from WLPO alone

fredrik-bakke:


source/CantorSchroederBernstein/CSB-WLPO.lagda
```diff
@@ -0,0 +1,286 @@
...
+open import UF.Embeddings
+open import UF.Equiv
+open import NotionsOfDecidability.Complemented
+open import NotionsOfDecidability.Decidable
+open import UF.Retracts
+
+\end{code}
+
+In this file we consider a generalization of the Cantor–Schröder–Bernstein (CSB)
+theorem only assuming the weak limited principle of omniscience (WLPO), based on
```
martinescardo: "only" is a stretch. You also assume that the embeddings are complemented. (And yes, I did see the discussion below.)

source/CantorSchroederBernstein/CSB-WLPO.lagda
```diff
@@ -0,0 +1,286 @@
...
+ map-construction : (x : X) → is-decidable (is-perfect-image f g x) → Y
+ map-construction x (inl γ) = inverse-on-perfect-image x γ
+ map-construction x (inr nγ) = f x
+
+ map-construction-CSB :
+  ((x : X) → is-decidable (is-perfect-image f g x)) → X → Y
+ map-construction-CSB D x = map-construction x (D x)
+
+ map-construction-is-left-cancellable :
+  (lc-f : left-cancellable f) →
```
martinescardo: The arrows are in the wrong place. They should be at the beginning of lines.

source/CantorSchroederBernstein/CSB-WLPO.lagda
```diff
@@ -0,0 +1,286 @@
...
+  {x x' : X}
+  (d : is-decidable (is-perfect-image f g x))
+  (d' : is-decidable (is-perfect-image f g x')) →
+  map-construction x d ＝ map-construction x' d' → x ＝ x'
+ map-construction-is-left-cancellable lc-f {x} {x'} (inl ρ) (inl ρ') p =
+  inverse-on-perfect-image-is-section x ρ ⁻¹
```
martinescardo: Please use chains of equations for readability.

source/CantorSchroederBernstein/CSB-WLPO.lagda
martinescardo: Fix this everywhere please.

source/CantorSchroederBernstein/CSB-WLPO.lagda
```diff
@@ -0,0 +1,286 @@
...
+  map-construction-is-left-cancellable lc-f (D x) (D x')
+
+\end{code}
+
+Computations with the construction.
+
+\begin{code}
+
+module _
+ {X : 𝓤 ̇ } {Y : 𝓥 ̇ } {f : X → Y} {g : Y → X}
```
martinescardo: See how we indent parameters in anonymous modules by looking at examples in TypeTopology.

source/CantorSchroederBernstein/CSB-WLPO.lagda
```diff
@@ -0,0 +1,286 @@
...
+ (¬¬elim-g : is-¬¬-stable-map g)
+ (lc-g : left-cancellable g)
+ (αf : is-¬¬-Compact'-map f {𝓤 ⊔ 𝓥})
+ where
+
+ compute-construction-on-perfect-image :
```
martinescardo: Can you explain the terminology "compute" here?

source/CantorSchroederBernstein/CSB-WLPO.lagda
```diff
@@ -0,0 +1,286 @@
...
+ compute-construction-on-not-perfect-image y nγ (inl v) =
+  𝟘-elim (nonperfect-fiber-of-not-perfect-image-is-not-perfect' αf ¬¬elim-g lc-g y nγ v)
+ compute-construction-on-not-perfect-image y nγ (inr _) =
+  compute-element-in-nonperfect-fiber-of-not-perfect-image' αf ¬¬elim-g lc-g y nγ
+\end{code}
+
+The construction is an equivalence.
+
+\begin{code}
+ inverse-construction :
```
martinescardo: Blank lines needed after `begin` and before `end` (everywhere).

source/CantorSchroederBernstein/CSB-WLPO.lagda
```diff
@@ -0,0 +1,286 @@
...
+ is-section-inverse-construction-CSB D y =
+   construction-is-retraction y
+    (D (g y))
+    (D (inverse-construction-CSB (D ∘ g) y))
+
+ map-construction-CSB-has-section
+  : (D : (x : X) → is-decidable (is-perfect-image f g x))
+  → has-section (map-construction-CSB D)
+ map-construction-CSB-has-section D =
+  ( inverse-construction-CSB (D ∘ g) ,
```
martinescardo: Space after bracket should be removed.

martinescardo: Everywhere.

source/CantorSchroederBernstein/CSB-WLPO.lagda
```diff
@@ -0,0 +1,286 @@
...
+  left-cancellable f →
+  (D : (x : X) → is-decidable (is-perfect-image f g x))
+  → is-equiv (map-construction-CSB D)
+ construction-CSB-is-equiv lc-f D =
+  lc-retractions-are-equivs (map-construction-CSB D)
+   (map-construction-CSB-is-left-cancellable lc-f D)
+   (map-construction-CSB-has-section D)
+
+ equiv-CSB :
+  left-cancellable f
```
martinescardo: The ":" should start in the next line (everywhere), unless you indent everything after the colon. The arrows should be indendent in the same column as the colon (in all cases).

source/CantorSchroederBernstein/CSB-WLPO.lagda
```diff
@@ -0,0 +1,286 @@
...
+
+\end{code}
+
+Note in particular that the above definition gives us a fully constructive
+version of König's argument:
+
+If f and g are such that
+
+ 1. g is left cancellable and ¬¬-stable,
+ 2. f is left cancellable and has ¬¬-compact fibers
```
martinescardo: Are there any *examples* of maps satisfying (1)-(3)?

source/CantorSchroederBernstein/CSB-WLPO.lagda
```diff
@@ -0,0 +1,286 @@
...
+ retract-CSB-assuming-WLPO cg emb-g αf βf =
+  retract-CSB
+   (λ y → ¬¬-elim (cg y))
+   (embeddings-are-lc g emb-g)
+   (αf)
+   (lemma cg emb-g βf)
+
+ equiv-CSB-assuming-WLPO : is-complemented-map g
+                         → is-embedding g
+                         -- The following three assumptions imply
```
martinescardo: Please write these comments outside the code block.

source/CantorSchroederBernstein/PerfectImages.lagda
```diff
@@ -0,0 +1,342 @@
...
+Perfect images
+--------------
+
+\begin{code}
+
+{-# OPTIONS --safe --without-K #-}
+
+module CantorSchroederBernstein.PerfectImages where
+
+
```
martinescardo: One blank line too many.

source/MLTT/Id.lagda
```diff
@@ -69,6 +69,14 @@ refl ∙' q = q
                  → p ∙' q ＝ p ∙ q
 ∙-agrees-with-∙' refl refl = refl
 
+right-unit-law-∙ : {X : 𝓤 ̇ } {x y : X} (p : x ＝ y)
```
martinescardo: These already exist in `UF.Base`.

source/TypeTopology/CompactTypes.lagda
```diff
@@ -1242,3 +1249,42 @@ Compact-gives-Compact' C A _ = C A
 
 TODO. (1) is-Compact' X ≃ is-compact X.
       (2) is-Compact' X is a retract of is-Compact X.
+
+Added by Fredrik Bakke on the 2nd of April 2025.
+
+Π-Compact types are closed under dependent sums.
```
martinescardo: "dependent sums" -> "sums".

martinescardo (changes requested):
Here is a first pass of review. I didn't indicate everything. Please try to format the whole code according to our TypeTopology conventions. I will review this again for technical contents after you make these changes. Thanks.

source/CantorSchroederBernstein/CSB-WLPO.lagda
fredrik-bakke: Apologies, my bad

source/CantorSchroederBernstein/CSB-WLPO.lagda
fredrik-bakke: If we do not assume WLPO you mean? I'll think about it, I'd love to know! Even if we drop the left cancellability condition on f we may still conclude that Y is a retract of X in fact.

source/CantorSchroederBernstein/CSB-WLPO.lagda
fredrik-bakke: I use the descriptor `compute` in the sense that the entry at hand _computes_ for us the value of [the relevant construction] when [restricted to the perfect image]. In this particular case the computation is done by producing a witness of equality to some simpler term, but I could've also figured to use it for an equivalence of types.

I suppose I imported this descriptor from agda-unimath, so apologies if it goes against conventions. Usually it replaces words like `eq`, `equals`, or `equiv`.

source/CantorSchroederBernstein/CSB-WLPO.lagda
martinescardo: I think a more blunt question is: is this theorem useful? Can it ever be applied? Can its hypothesis be fulfilled in an interesting way?

source/CantorSchroederBernstein/CSB-WLPO.lagda
martinescardo: I don't like this use of "compute". Agda computes. But the definitions don't compute anything. They just define things. Can you think of a better name?

fredrik-bakke:
Hi again, I'm back from vacation now and picking back up on this.

source/CantorSchroederBernstein/CSB-WLPO.lagda
fredrik-bakke: Do you mean aside from in obtaining the main result of this file? I realize the explanation was not sufficiently clear that this characterization is immediately applied under the assumption of WLPO. I've adjusted it now.

You would also be right to point out that this contribution does not discuss any examples, hence the statement that the main result is a _proper_ generalization is not demonstrated.

source/CantorSchroederBernstein/CSB-WLPO.lagda
fredrik-bakke: While thinking about this I realized that I had the order of words wrong on the names in this file. I've updated them now to hopefully be more aligned with the naming style of TypeTopology.

source/CantorSchroederBernstein/CSB-WLPO.lagda
fredrik-bakke: I've updated the introduction to remove the word "proper", until an example is given.

source/CantorSchroederBernstein/CSB-WLPO.lagda
```diff
@@ -0,0 +1,360 @@
...
+References
+──────────
+ - König, 1906. Sur la théorie des ensembles.
+   https://gallica.bnf.fr/ark:/12148/bpt6k30977.image.f110.langEN
+
+ - Pradic & Brown, 2022. Cantor-Bernstein implies Excluded Middle.
+   https://arxiv.org/abs/1904.09193
+
+ - Bakke, 2025. The Cantor–Schröder–Bernstein theorem in ∞-Topoi.
+   https://hott-uf.github.io/2025/slides/Bakke.pdf (slides)
```
martinescardo, line 372: Can you add two references, please (1) my own Agda code in TypeTopology, (2) my published paper resulting from this? Thanks.

fredrik-bakke: Of course! Thanks for mentioning it.

PR #361, opened 2025-07-03
Simpler proof that Nat is a set without universes, plus same for Lists

ecavallo:
I just put the proof for lists in the same file. Should the file be renamed?


martinescardo (approved):
Nice! Thanks again. 

Regarding the question whether the file should be renamed, I don't have strong views either way. Feel free to decide.

PR #362, opened 2025-07-08
Prove equivalence of untruncated at-most-two lemma for Aut Omega and EM for Aut Omega

jasoncarr0:
The lemma used to prove this probably could serve to be moved somewhere else. I think a section for different notions of cardinality comparisons would be nice.

It implies also that a truncated form of the has-a-repetition would be very reasonable.

The proof is novel as far as I know, although the idea is inspired by Diaconescu's theorem. When we have a function whose domain is extensional, we can force it to make commitments on things which we later learn are equal.

### DIFF
--- a/source/Higgs/UntruncatedAtMostTwo.lagda
+++ b/source/Higgs/UntruncatedAtMostTwo.lagda
@@ -130,11 +130,11 @@ almost-constantly-is-constant {_} {_} {X} {Y} X-discrete x' y _ refl = dfunext f
 
 
 at-most-discrete-gives-discrete
-  : {𝓤 𝓥 : Universe}
-  → (X : 𝓤 ̇) (Y : 𝓥 ̇)
-  → is-discrete X
-  → ((f : X → Y) → f has-a-repetition)
-  → is-discrete Y
+ : {𝓤 𝓥 : Universe}
+ → (X : 𝓤 ̇) (Y : 𝓥 ̇)
+ → is-discrete X
+ → ((f : X → Y) → f has-a-repetition)
+ → is-discrete Y
 at-most-discrete-gives-discrete X Y X-discrete f-ph y y' = V VI
  where
 
@@ -145,15 +145,15 @@ at-most-discrete-gives-discrete X Y X-discrete f-ph y y' = V VI
     (f-ph f)
 
   repeat-is-repeat : (f : X → Y)
-                   → let (x , x') = repeat-indices f in
-                   f x ＝ f x'
+                   → let (x , x') = repeat-indices f
+                     in f x ＝ f x'
   repeat-is-repeat f =
     let (x , x' , _ , pf) = f-ph f
     in pf
 
   repeat-distinct : (f : X → Y)
-                → let (x , x') = repeat-indices f in
-                x ≠ x'
+                  → let (x , x') = repeat-indices f
+                    in x ≠ x'
   repeat-distinct f =
     let (x , x' , pf , _) = f-ph f
     in pf
@@ -165,7 +165,7 @@ at-most-discrete-gives-discrete X Y X-discrete f-ph y y' = V VI
 
   I : y ＝ y' → ix₁ ＝ ix₂
   I e = ap repeat-indices
-    (almost-constantly-is-constant X-discrete (pr₁ ix₁) y y' e)
+           (almost-constantly-is-constant X-discrete (pr₁ ix₁) y y' e)
 
   II : (x : X)
      → (pr₁ ix₁ ≠ x)
@@ -177,14 +177,14 @@ at-most-discrete-gives-discrete X Y X-discrete f-ph y y' = V VI
 
   IV : ix₁ ＝ ix₂ → y ＝ y'
   IV e =
-    y            ＝⟨ II (pr₂ ix₁) (repeat-distinct f₁) ⟩
-    f₂ (pr₂ ix₁) ＝⟨ ap (f₂ ∘ pr₂) e ⟩
-    f₂ (pr₂ ix₂) ＝⟨ refl ⟩
-    f₂ (pr₂ (repeat-indices f₂)) ＝⟨ repeat-is-repeat f₂ ⁻¹ ⟩
-    f₂ (pr₁ (repeat-indices f₂)) ＝⟨ refl ⟩
-    f₂ (pr₁ ix₂) ＝⟨ ap (f₂ ∘ pr₁) (e ⁻¹) ⟩
-    f₂ (pr₁ ix₁) ＝⟨ III ⟩
-    y' ∎
+   y                            ＝⟨ II (pr₂ ix₁) (repeat-distinct f₁) ⟩
+   f₂ (pr₂ ix₁)                 ＝⟨ ap (f₂ ∘ pr₂) e ⟩
+   f₂ (pr₂ ix₂)                 ＝⟨ refl ⟩
+   f₂ (pr₂ (repeat-indices f₂)) ＝⟨ repeat-is-repeat f₂ ⁻¹ ⟩
+   f₂ (pr₁ (repeat-indices f₂)) ＝⟨ refl ⟩
+   f₂ (pr₁ ix₂)                 ＝⟨ ap (f₂ ∘ pr₁) (e ⁻¹) ⟩
+   f₂ (pr₁ ix₁)                 ＝⟨ III ⟩
+   y'                           ∎
 
   V : is-decidable (ix₁ ＝ ix₂) → is-decidable (y ＝ y')
   V (inl e) = inl (IV e)
@@ -200,11 +200,11 @@ We may write the untruncated form of the at-most-2 lemma in this form
 \begin{code} 
 
 at-most-two-is-pigeonhole
-  : {𝓤 : Universe}
-  → {X : 𝓤 ̇}
-  → ((x y z : X) → (z ＝ x) + (x ＝ y) + (y ＝ z))
-  → (f : Fin 3 → X)
-  → f has-a-repetition
+ : {𝓤 : Universe}
+ → {X : 𝓤 ̇}
+ → ((x y z : X) → (z ＝ x) + (x ＝ y) + (y ＝ z))
+ → (f : Fin 3 → X)
+ → f has-a-repetition
 at-most-two-is-pigeonhole at-most-2 f = II I
  where
   v1 v2 v3 : Fin 3
@@ -244,11 +244,10 @@ at-most-two-is-pigeonhole at-most-2 f = II I
   II (inr (inr e23)) = ( v2 , v3 , v2-not-3 , e23 )
 
 aut-Ω-discrete-has-em
-  : is-discrete (Aut Ω)
-  → (𝕗 : Aut Ω)
-  → (𝕗 ＝ 𝕚𝕕) + (𝕗 ≠ 𝕚𝕕)
-aut-Ω-discrete-has-em aut-disc 𝕗 =
-  aut-disc 𝕗 𝕚𝕕
+ : is-discrete (Aut Ω)
+ → (𝕗 : Aut Ω)
+ → (𝕗 ＝ 𝕚𝕕) + (𝕗 ≠ 𝕚𝕕)
+aut-Ω-discrete-has-em aut-disc 𝕗 = aut-disc 𝕗 𝕚𝕕
 
 untruncated-at-most-two-iff-em
  : ((f g h : Aut Ω) → (h ＝ f) + (f ＝ g) + (g ＝ h))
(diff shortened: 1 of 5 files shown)


source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+The core method is to note that the pidgeonhole function must respect equality
+of functions, so we produce a pair of functions, where equality of their results
+holds if and only if the two elements of the codomain are equal.
+
+\begin{code}
+
+constantly : {𝓤 𝓥 : Universe}
+           → {X : 𝓤 ̇ } {Y : 𝓥 ̇ }
+           → Y → X → Y
+constantly const-y _ = const-y
```
martinescardo: Please call this `y` rather than `const-y`.

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+
+almost-constantly : {𝓤 𝓥 : Universe}
+                  → {X : 𝓤 ̇ } {Y : 𝓥 ̇ }
+                  → is-discrete X
+                  → X → Y → Y → X → Y
+almost-constantly X-discrete pt-x pt-y const-y x' =
```
martinescardo: Please call these `x` and `y` and `y'` rather than what you have. The current names are confusing.

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+ → (y y' : Y)
+ → y ＝ y'
+ → constantly y ＝ almost-constantly X-discrete x y' y
+almost-constantly-is-constant {_} {_} {X} {Y} X-discrete x y _ refl = dfunext fe I
+ where
+  I : (x' : X)
+    → constantly y x' ＝ almost-constantly X-discrete x y y x'
+  -- possibly rewrite this with dep-cases?
+  -- it's more obvious and spartan but much more awkward
+  I x' with X-discrete x' x
```
martinescardo: We don't use `with` in TypeTopology. There are a few `with`s left, but the intension is to get rid of all of them. 

Reasons (among others):
(1) It is difficult to prove things about things defined with `with`.
(2) It has a rather unclear semantics, and we try to use a fragment of Agda that amounts to MLTT.

martinescardo: Use e.g. `Cases` or `dep-Cases` instead, or write a helper function (I often do the latter).

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+  ... | inr _ = refl
+
+at-most-discrete-gives-discrete
+  : {𝓤 𝓥 : Universe}
+  → (X : 𝓤 ̇) (Y : 𝓥 ̇)
+  → is-discrete X
+  → ((f : X → Y) → f has-a-repetition)
+  → is-discrete Y
+at-most-discrete-gives-discrete X Y X-discrete f-ph y y' = V VI
+ where
```
martinescardo, line 139: What is `ph` meant to mean?

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+
+  repeat-indices : (X → Y)
+                 → X × X
+  repeat-indices f =
+    (λ (x , x' , _) → (x , x'))
+    (f-ph f)
+
+  repeat-is-repeat : (f : X → Y)
+                   → let (x , x') = repeat-indices f in
```
martinescardo: Please indent `in` below `let` and then write the term in in the same line as `in`.

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+                   f x ＝ f x'
+  repeat-is-repeat f =
+    let (x , x' , _ , pf) = f-ph f
+    in pf
+
+  repeat-distinct : (f : X → Y)
+                → let (x , x') = repeat-indices f in
```
martinescardo: Bad indentation, *and* same problem with `let`.

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+    in pf
+
+  f₁ = constantly y
+  ix₁ = repeat-indices f₁
+  f₂ = almost-constantly X-discrete (pr₁ ix₁) y' y
+  ix₂ = repeat-indices f₂
+
+  I : y ＝ y' → ix₁ ＝ ix₂
+  I e = ap repeat-indices
+    (almost-constantly-is-constant X-discrete (pr₁ ix₁) y y' e)
```
martinescardo: Indentation.

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+
+  II : (x : X)
+     → (pr₁ ix₁ ≠ x)
+     → y ＝ f₂ x
+  II x ne with X-discrete x (pr₁ ix₁)
```
martinescardo: Please get rid of all `with`s.

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+  ... | inr _ = refl
+
+  III : f₂ (pr₁ ix₁) ＝ y'
+  III with X-discrete (pr₁ ix₁) (pr₁ ix₁)
+  ... | inl e = refl
+  ... | inr ne = 𝟘-elim (ne refl)
+
+  IV : ix₁ ＝ ix₂ → y ＝ y'
+  IV e =
+    y            ＝⟨ II (pr₂ ix₁) (repeat-distinct f₁) ⟩
```
martinescardo: Align the equality symbols and the end of equality proof symbol.

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+  v2 = inl (inr ⋆)
+  v3 = inl (inl (inr ⋆))
+
+  true-when : Fin 3
+            → Fin 3
+            → 𝓤 ⁺ ̇
+  true-when (inr _) (inr _) = 𝟙
+  true-when (inl (inr _)) (inl (inr _)) = 𝟙
+  true-when (inl (inl (inr _))) (inl (inl (inr _))) = 𝟙
+  {-# CATCHALL #-}
```
martinescardo: I am not sure what `{-# CATCHALL #-}` is doing.

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+  I = at-most-2 (f v1) (f v2) (f v3)
+
+  II : (f v3 ＝ f v1) + (f v1 ＝ f v2) + (f v2 ＝ f v3)
+     → f has-a-repetition
+  II (inl e31)       = ( v3 , v1 , v3-not-1 , e31 )
+  II (inr (inl e12)) = ( v1 , v2 , v1-not-2 , e12 )
+  II (inr (inr e23)) = ( v2 , v3 , v2-not-3 , e23 )
+
+aut-Ω-discrete-has-em
+  : is-discrete (Aut Ω)
```
martinescardo: Indentation is by one space (please check everywhere).

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,244 @@
...
+untruncated-at-most-two-iff-em = (FW , BW)
+ where
+  FW : ((f g h : Aut Ω) → (h ＝ f) + (f ＝ g) + (g ＝ h))
+     → ((𝕗 : Aut Ω) → (𝕗 ＝ 𝕚𝕕) + (𝕗 ≠ 𝕚𝕕))
+  FW at-most-two = aut-Ω-discrete-has-em
+    (at-most-discrete-gives-discrete
+      (Fin 3) (Aut Ω)
+      Fin-is-discrete
+      (at-most-two-is-pigeonhole at-most-two)
+    )
```
martinescardo: Please close the bracket in the previous line.

martinescardo (changes requested):
Thanks for your contribution, and sorry it took two weeks for me to look at this.

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: Otherwise Agda was rejecting it for me when I built, or at least giving a warning, for fallthrough cases.
Since that line is catching several different pairs of constructors

martinescardo: Please fix this by writing the missing cases. Are you using case-split on emacs (`ctrl-c-ctrl-c`) to generate the cases? Using this, the `{-# CATCHALL #-}` should not be needed. Thanks.

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: I think I originally had x x' y y', but no matter which way these are oriented I find it confusing and errorprone. Changed nonetheless

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: I fixed this, although it made the proofs significantly more verbose

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: Pidgeonhole, not sure of a good name for it. But the principle that every function from X into Y has a repeat in some Y.

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: Fixed

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: Fixed

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: Not sure what's the best here, I've moved this instance onto one line. In some other places I guessed you wanted one space?

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: Done as noted above

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: Fixed

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: Okay, I've done this, it's harder to read now in my opinion.
The issue was just a warning from Agda saying to the effect that the syntax didn't match what would reduce definitionally.

The automatic case-split here was quite tedious since it makes 9 cases and requires splitting each time.

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: Hopefully fixed everywhere

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: Done

jasoncarr0:
One minor note about the location and relation to finiteness in general.
This is (slightly more general than) the converse of https://github.com/martinescardo/TypeTopology/blob/master/source/Fin/Pigeonhole.lagda#L172

I.e. we have
pigeonhole-principle-iff : (Y: 𝓤 ̇ ) (m : ℕ) →
   ((f : Fin m → Y) → f has-a-repetition) ↔
   (n : ℕ , n < m × Y ≃ Fin n)

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,290 @@
...
+                     → (y' y : Y)
+                     → almost-constantly X-discrete x y' y x ＝ y'
+almost-constantly-eq X-discrete x y' y =
+ almost-constantly X-discrete x y' y x               ＝⟨ refl ⟩
+ almost-constantly-inner x y' y x (X-discrete x x)   ＝⟨ I ⟩
+ almost-constantly-inner x y' y x (inl refl)         ＝⟨ refl ⟩
+ y'                                                  ∎
+ where
+  I : almost-constantly-inner x y' y x (X-discrete x x) ＝
+      almost-constantly-inner x y' y x (inl refl)
```
martinescardo: Our (probably unwritten) convention for `=` is the same as with `→`. When we break a line, we don't leave the symbol last, but instead first at the next line.

source/Higgs/UntruncatedAtMostTwo.lagda
```diff
@@ -0,0 +1,290 @@
...
+                      → (y' y : Y)
+                      → (x' ≠ x)
+                      → almost-constantly X-discrete x' y' y x ＝ y
+almost-constantly-neq X-discrete x' x y' y ν =
+ almost-constantly X-discrete x' y' y x               ＝⟨ refl ⟩
+ almost-constantly-inner x' y' y x (X-discrete x' x)  ＝⟨ I ⟩
+ almost-constantly-inner x' y' y x (inr ν)            ＝⟨ refl ⟩
+ y                                                    ∎
+ where
+  I :  almost-constantly-inner x' y' y x (X-discrete x' x) ＝
```
martinescardo: Same here. Please check everywhere.

source/Higgs/UntruncatedAtMostTwo.lagda
jasoncarr0: Sorry for the slow response, this should be fixed by the conventions I saw in the repository

martinescardo (approved):
Thanks.

PR #363, opened 2025-07-16
More on ordinal arithmetic

tomdjong:
The main addition is the file `Ordinals/BoundedOperations.lagda`. See the comments throughout that file for a description.

En passant, I also show that both
* ((α β : Ordinal 𝓤) → 𝟘ₒ ⊲ α → β ⊴ α ×ₒ β) and
* ((α β : Ordinal 𝓤) → β ⊴ α +ₒ β)

are equivalent to EM 𝓤.

NB. We already had that EM 𝓤 is equivalent to ((α β : Ordinal 𝓤) → 𝟙ₒ ⊲ α → β ⊴ α ^ₒ β).


martinescardo (approved):
Nice, and interesting.

PR #364, opened 2025-07-22
First draft of joins of maps. Investigating join construction soon.

IanRay11:
Join of maps is defined. Soon I will investigate the join construction paper and see the extent to which the flattening lemma is required.

### DIFF
--- a/source/UF/JoinConstruction.lagda
+++ b/source/UF/JoinConstruction.lagda
@@ -112,7 +112,7 @@ approximations.
 
 \end{code}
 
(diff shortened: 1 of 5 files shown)


martinescardo:
What is the status of this PR. @IanRay11 ?

IanRay11:
I have intentions of returning to this but it's not a priority and will likely require quite a bit of effort to work out what results for pushouts I need. I don't think I need full descent theorems but something of that flavor for certain.

PR #367, opened 2025-08-13
Record asserting the existence of pushouts of arbitrary spans

IanRay11:
…olymorphically


tomdjong:
@IanRay11 I think that the following line is a little unclear.
> We now provide a record that allows the existence of pushouts to be assumed polymorphically.

How about this instead?
> We now provide a record that asserts the existence of pushouts of arbitrary spans.

PR #369, opened 2025-08-19
Subtypes are locally small (wrt reference universe)

IanRay11:
I also removed an inlined equivalence argument from a special case of my lemma, since it was later added to the library.


tomdjong:
Nice! Could you add your result and `small-implies-locally-small` both up so that they follow right after `subtype-is-locally-small` and go before the set replacement stuff, please?

IanRay11:
@tomdjong this required moving up the introduction of locally_small and I gave a brief comment about the generalization even though this wasn't my addition (hope that is fine).

PR #370, opened 2025-08-21
Forms of replacement and their logical equivalence

IanRay11:
…alence

### DIFF
--- a/source/UF/Replacement.lagda
+++ b/source/UF/Replacement.lagda
@@ -9,8 +9,8 @@ from the set theoretic axiom of replacement, where the set vs class distinction
 provides a notion of size. In type theory, this statement may be assumed or
 proven depending on the context. We also mention that the file UF/Size.lagda
 defines 'set replacement' which restricts type replacement to those maps whose
-codomain is a set (in the sense of univalent foundations). We will drop the term
-'axiom' and call it 'type replacement' or simply 'replacement' when there is no
+codomain is a set (in the sense of univalent foundations). We will drop the word
+'axiom' and write 'type replacement', or simply 'replacement' when there is no 
 risk of confusion with set replacement.
 
 The statement of type replacement is as follows:


source/UF/Replacement.lagda
```diff
@@ -0,0 +1,107 @@
...
+
+The type theoretic axiom of replacement (section 24.5 of Introduction to
+Homotopy Type Theory by Egbert Rijke:
+https://ulrikbuchholtz.dk/hott1920/hott-intro.pdf)
+is a statement about the size of the image of a function when certain smallness
+assumptions are imposed on the domain and codomain. The nomenclature is derived
+from the set theoretic axiom of replacement, where 'sethood' provides a notion
+of size. In type theory, this statement may be assumed or proven depending on
+the context. Thus, we will call it 'type replacement' or simply 'replacement'
+(when there is no risk of confusion with set replacement).
```
tomdjong: There should also be a reference to Set Replacement, as defined in `UF/Size`, which is where we restrict to situations where the codomain is a set.

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,107 @@
...
+
+The statement of type replacement is as follows:
+The image of a map f : A → X, from a small type A to a locally small type X, is
+itself small.
+
+Note: some authors use the term 'essentially small' for what the TypeTopology
```
tomdjong, line 20: Typo: the TypeTopology

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,107 @@
...
+refers to as simply 'small'. Although, often times it is desirable to explicitly
+state the universe with which a type is small relative to, as you will see in
```
tomdjong: *with which a type is small relative to* — Can you rephrase this in a simpler way please?

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,107 @@
...
+the code below.
+
+Type replacement is provable in the presence of a certain class of higher
+inductive types (HITs). In particular, "The Join Construction" by Egbert Rijke
+(https://arxiv.org/abs/1701.07538.) provides a construction that allows a proof
+of type replacement in the presence of 'graph quotients'. More conservativly one
```
tomdjong: Typo: conservativly

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,107 @@
...
+itself small.
+
+Note: some authors use the term 'essentially small' for what the TypeTopology
+refers to as simply 'small'. Although, often times it is desirable to explicitly
+state the universe with which a type is small relative to, as you will see in
+the code below.
+
+Type replacement is provable in the presence of a certain class of higher
+inductive types (HITs). In particular, "The Join Construction" by Egbert Rijke
+(https://arxiv.org/abs/1701.07538.) provides a construction that allows a proof
```
tomdjong: Bit convoluted: *a construction that allows a proof*.

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,107 @@
...
+of type replacement in the presence of 'graph quotients'. More conservativly one
+may carry out this construction merely with pushouts (in fact, one only requires
+the join of maps and sequential colimits, which are instances of pushouts).
+This route is actively being explored in other TypeTopology files.
+
+It is worth noting that the status of type replacement in the hierarchy of HIT
```
tomdjong: What does hierarchy of HIT strength mean when type replacement is not itself a HIT?

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,107 @@
...
+
+\end{code}
+
+We begin by giving a precise statement of type replacement.
+
+\begin{code}
+
+module _ {𝓥 : Universe} where
+
+ Replacement : {𝓤 𝓦 : Universe} → (𝓥 ⁺) ⊔ (𝓤 ⁺) ⊔ (𝓦 ⁺) ̇
```
tomdjong: You can replace `(𝓥 ⁺) ⊔ (𝓤 ⁺) ⊔ (𝓦 ⁺)` by `(𝓥 ⊔ 𝓤 ⊔ 𝓦) ⁺`.

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,107 @@
...
+                             → Replacement {𝓤} {𝓦} → Replacement' {𝓤} {𝓦}
+ Replacement-to-Replacement' 𝓡 {_} {X} f A-small X-loc-small f-surj
+  = smallness-closed-under-≃ I II
+  where
+   I : image f is 𝓥 small
+   I = 𝓡 f A-small X-loc-small
+   II : image f ≃ X
+   II = surjection-≃-image f f-surj
+
+ open propositional-truncations-exist pt
```
tomdjong: This should be done at the top somewhere.

tomdjong (changes requested):
Thanks!

source/UF/Replacement.lagda
IanRay11: I'll try to give a more precise rewording.

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,108 @@
...
+
+The type theoretic axiom of replacement (section 24.5 of Introduction to
+Homotopy Type Theory by Egbert Rijke:
+https://ulrikbuchholtz.dk/hott1920/hott-intro.pdf)
+is a statement about the size of the image of a function when certain smallness
+assumptions are imposed on the domain and codomain. The nomenclature is derived
+from the set theoretic axiom of replacement, where 'sethood' provides a notion
+of size. In type theory, this statement may be assumed or proven depending on
+the context. Thus, we will call it 'type replacement' or simply 'replacement'
+when there is no risk of confusion with set replacement (see Size.lagda).
```
tomdjong: Sorry but now the comment is more confusing than before as it currently suggests that `set` in `set replacement (see Size.lagda)` corresponds to *size* (set vs class), but it doesn't, it refers to a truncation level.

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,108 @@
...
+inductive types (HITs). In particular, "The Join Construction" by Egbert Rijke
+(https://arxiv.org/abs/1701.07538.) uses 'graph quotients' to give an
+alternative construction of the image where type replacement can be proven.
+More conservatively one may carry out this construction merely with pushouts
+(in fact, one only requires the join of maps and sequential colimits, which are
+instances of pushouts). This route is actively being explored in other
+TypeTopology files.
+
+It is worth noting that the status of type replacement's strength relative to
+other HITs is not completely understood, but it appears to be weaker than the
```
tomdjong: Remove "other" as type replacement itself is not a HIT. Maybe also rephrase to "relative to the strength of adding certain HITs".

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,108 @@
...
+assumption that pushouts exist (this observation will follow from a forthcoming
+write up by Reid Barton). In light of this, it is reasonable to explore type
+replacement and use it as an independent assumption when neccesary.
+
+\begin{code}
+
+{-# OPTIONS --safe --without-K #-}
+
+
```
tomdjong: Remove one of the double blank lines.

tomdjong (changes requested):

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,110 @@
+Ian Ray, 17th August 2025.
+
+The type theoretic axiom of replacement (section 24.5 of Introduction to
+Homotopy Type Theory by Egbert Rijke:
+https://ulrikbuchholtz.dk/hott1920/hott-intro.pdf)
+is a statement about the size of the image of a function when certain smallness
```
tomdjong, line 6: Sorry, but I'm still not 100% happy with this paragraph. 
1. > In type theory, this statement may be assumed or proven depending on the context. Thus, we will call it 'type replacement' or simply 'replacement'.

   Why "Thus", i.e., why does the first claim mean that those names are suitable?

1. Maybe the following is clearer.
   > The nomenclature is derived from the set theoretic axiom of replacement, where the set vs class distinction provides a notion of size.

1. Maybe the last bit can be replaced by the following.
   > We also mention that the file UF/Size.lagda defines 'set replacement' which restricts type replacement to those maps whose codomain is a set in the sense of univalent foundations. This set replacement principle is equivalent to having small set quotients, see Quotient/index.lagda.

IanRay11:
I addressed 2) and 3).

source/UF/Replacement.lagda
```diff
@@ -0,0 +1,112 @@
...
+assumptions are imposed on the domain and codomain. The nomenclature is derived
+from the set theoretic axiom of replacement, where the set vs class distinction
+provides a notion of size. In type theory, this statement may be assumed or
+proven depending on the context. We will drop the term 'axiom' and call it
```
tomdjong: The explanation for dropping the word 'axiom' is clear, thanks! 
What is still confusing is that when you write *no risk of confusion with set replacement*, it is not clear what *set replacement* refers to. Is it the set theoretic axiom of replacement? Or is the set replacement from `UF/Size.lagda` (which has not been introduced yet at this point!)?

IanRay11:
What if I put that sentence last? Also, I think this part "This set replacement principle is equivalent to having small set quotients, see Quotient/index.lagda." is not neccesary here, what do you think? @tomdjong

IanRay11 (commented):
NA

source/UF/Replacement.lagda
```diff
@@ -7,12 +7,11 @@ is a statement about the size of the image of a function when certain smallness
...
-'type replacement' or simply 'replacement' when there is no risk of confusion
-with set replacement. We also mention that the file UF/Size.lagda defines 'set
-replacement' which restricts type replacement to those maps whose codomain is
-a set in the sense of univalent foundations. This set replacement principle is
-equivalent to having small set quotients, see Quotient/index.lagda.
+proven depending on the context. We also mention that the file UF/Size.lagda
+defines 'set replacement' which restricts type replacement to those maps whose
+codomain is a set (in the sense of univalent foundations). We will drop the term
+'axiom' and call it 'type replacement' or simply 'replacement' when there is no
+risk of confusion with set replacement.
```
tomdjong, line 14: This is much better, but I suggest a further small revision.
```quote
We will drop the word 'axiom' and write 'type replacement', or simply 'replacement' when there is no 
risk of confusion with set replacement.
```

IanRay11: test

PR #375, opened 2025-08-28
Updating Size-TruncConn file with more coherent treatment of Replacement

IanRay11:
Here I define a third version of replacement stated in terms of truncation levels (that is, -1 connected maps as opposed to surjections, etc.) and iterated local smallness. This version of replacement is again proven to be logically equivalent to the surjection version. We assume this version and univalence when neccesary (in unnamed modules).

### DIFF
--- a/source/UF/Size-TruncatedConnected.lagda
+++ b/source/UF/Size-TruncatedConnected.lagda
@@ -142,7 +142,7 @@ We note that while the name for this result is due to its 'apparent' similarity
 to the axiom of replacement from set theory, but replacement in type theory is
 quite a modest assumption in comparison. Indeed, type replacement follows from
 the existence of pushouts (see UF/Replacement for further discussion on the
-modest strength of Replacement) as demonstrated by "The Join Construction" by
+modest strength of replacement) as demonstrated by "The Join Construction" by
 Egbert Rijke (https://arxiv.org/abs/1701.07538).
 
 In the interest of keeping the present work self-contained and requiring only


source/UF/Size-TruncatedConnected.lagda
```diff
@@ -44,13 +46,14 @@ private
...
 \end{code}
 
-We begin by giving some definitions from [1] and proving important properties about
-them. We have fixed the universe parameter 𝓥 and use it as our point of reference
-for 'smallness'. Also, note that in what follows we only use univalence locally
-as some critical results hold in its absence.
+We begin by giving some definitions from [1] and proving important properties
+about them. We have fixed the universe parameter 𝓥 and use it as our point of
+reference for 'smallness'. Also, note that in what follows we will explicilty
+assume univalence as some critical results hold in its absence.
```
tomdjong: Typo: `explicilty`. (Could you please run a spell checker on the whole file.)
This new phrasing is confusing, because the second part (`absence`) of the sentence (`some critical results hold in its absence`) seems to be in conflict with the assumption. I suppose that, like in the old formulation, the point is to make clear that you only assume it where needed, rather than assuming it globally.

IanRay11:
@martinescardo @tomdjong I have updated comments in both of these files. Where the comment in the replacement file is carrying most of the weight. This seemed best to me but I could be wrong.

source/UF/Replacement.lagda
```diff
@@ -21,21 +22,26 @@ Note: some authors use the term 'essentially small' for what the TypeTopology
...
-(https://arxiv.org/abs/1701.07538.) uses 'graph quotients' to give an
-alternative construction of the image where type replacement can be proven.
-More conservatively one may carry out this construction merely with pushouts
-(in fact, one only requires the join of maps and sequential colimits, which are
-instances of pushouts). This route is actively being explored in other
-TypeTopology files.
-
-It is worth noting that the status of type replacement's strength relative to
+The reader may be cautious about the strength of type replacement due to its
+'apparent' connection with set theories axiom of replacement; the latter is a
```
tomdjong: `set theories axiom of replacement` -> `set theory's axiom of replacement`?
Maybe clearer: `the axiom of replacement in set theories`.

source/UF/Replacement.lagda
```diff
@@ -21,21 +22,26 @@ Note: some authors use the term 'essentially small' for what the TypeTopology
...
+rather strong assumption critical for more advanced set theoretic results. In
+contrast, type replacement is a rather modest assumption. Indeed it follows
+from the existence of a restricted class of Higher Inductive Types (HITs). For
+details see "The Join Construction" by Egbert Rijke
+(https://arxiv.org/abs/1701.07538.) where 'graph quotients' are used to give an
```
tomdjong: `.)` -> `).`

source/UF/Replacement.lagda
```diff
@@ -21,21 +22,26 @@ Note: some authors use the term 'essentially small' for what the TypeTopology
...
+alternative construction of the image; from which type replacement can be
+proven. More conservatively, one may carry out this construction merely with
```
tomdjong: `More conservatively` — I don't think this is accurate. As explained at the bottom of page 2 of Egbert's article, we can get pushouts from graph quotients and vice versa.

source/UF/Size-TruncatedConnected.lagda
```diff
@@ -102,103 +107,115 @@ local-smallness-is-closed-under-Σ : {X : 𝓤 ̇ } {Y : X → 𝓦 ̇ } {n : 
...
 Many of the results in [1] follow from a fact that some call the type theoretic
 axiom of replacement. It states that given a small type A, a locally small type
-X and a function f : A → X the image of f is small. In the paper 'the join
-construction' by Egbert Rijke, the axiom of replacement is shown to follow from
-the join construction. Currently, the join construction and the derivation of the
-axiom of replacement are not implemented in the TypeTopology library. We will
-state a more convenient but equivalent form of replacement.
+X and a function f : A → X the image of f is small.
+
+We note that while the name for this result is due to its 'apparent' similarity
```
tomdjong, line 141: I don't think it's grammatically correct to end the sentence before the second part of the *while*-clause.

source/UF/Size-TruncatedConnected.lagda
```diff
@@ -102,103 +107,115 @@ local-smallness-is-closed-under-Σ : {X : 𝓤 ̇ } {Y : X → 𝓦 ̇ } {n : 
...
+to the axiom of replacement from set theory. Replacement in type theory is
+quite a modest assumption in comparison. Indeed, type replacement follows from
+the existence of pushouts (see the TypeTopology file on Replacement for further
```
tomdjong: Clearer: `TypeTopology file on Replacement` -> `UF/Replacement`

source/UF/Replacement.lagda
IanRay11: How about 'the axiom of replacement of set theory'.

source/UF/Replacement.lagda
IanRay11: Oh thanks!

source/UF/Replacement.lagda
tomdjong: Not very pretty because of the double `of`, but sure.

source/UF/Replacement.lagda
```diff
@@ -23,18 +23,18 @@ library refers to as simply 'small'. Additionally, we often wish to consider
...
+(https://arxiv.org/abs/1701.07538) where 'graph quotients' are used to give an
 alternative construction of the image; from which type replacement can be
-proven. More conservatively, one may carry out this construction merely with
-pushouts (in fact, one only requires the join of maps and sequential colimits,
-which are instances of pushouts). This route is being actively explored in other
-TypeTopology files. Additionally, set replacement (see above) is equivalent to
-the existence of set quotients, another modest assumption.   
+proven. Additionally, one may carry out this construction simply with pushouts
+(which can be shown to be equivalent to graph quotients). The join construction
+via pushouts is being actively explored in other TypeTopology files.
```
tomdjong: Sorry for yet another suggestion, but could you specify which files?

source/UF/Replacement.lagda
IanRay11: I did consider that but '... in set theories' sounded stranger. That could just be preference.

source/UF/Replacement.lagda
IanRay11: The problem is no of those files are public yet.

tomdjong: Then say *...explored in not yet publicly available files*.

martinescardo: Thanks, @IanRay11 for adding the comments I requested and @tomdjong for improving them.

PR #384, opened 2025-10-14
Basic facts about exponentiating natural numbers and their order

tomdjong:


source/Naturals/Order.lagda
```diff
@@ -938,3 +951,43 @@ double-reflects-< {zero} {succ y} _ = ⋆
...
+
+\begin{code}
+
+open import Naturals.Exponentiation
+
+exponential-positive-if-base-positive : (n m : ℕ) → 0 < n → 0 <ℕ n ℕ^ m
+exponential-positive-if-base-positive n zero _ = ⋆
+exponential-positive-if-base-positive n@(succ n') (succ m) l = II
+ where
+  IH : 0 <ℕ (n ℕ^ m)
```
martinescardo: Could we use the notation `<` rather than `<ℕ` here and other places (also for `≤ℕ`). Like is done is lines 809-812, for example.

martinescardo: Also `ℕ^` could be replaced by `^` e.g. with suitable use of `renaming` and/or `hiding` when importing.

tomdjong: Will do!

PR #393, opened 2025-11-20
Add syntax for definitional equational/equivalence reasoning steps.

cangiuli:
Following some discussion with @martinescardo both privately and on [this gist](https://gist.github.com/cangiuli/f6aa6fcc6d67a727c8a01ac0b06b997c), I have added special syntax for reflexive steps in equational (and equivalence) reasoning chains that are made explicit for reasons of readability. As described in the above gist, these new combinators do not insert any unnecessary reflexive paths/equivalences into the chain, in contrast to the standard equational/equivalence combinator (in many but not all cases).

As this is my first time contributing to TypeTopology, please let me know if I am not following any of your conventions correctly.

I also want to note that there are a few things that one might want that are not (yet?) included in this PR, notably (1) explicitly including the rationale for these combinators in the library (e.g., by committing a version of the above gist), or (2) changing any of the current code to actually use these combinators. As discussed in the comments of the above gist, a simple `grep` reveals at least 991 occurrences of reflexive equational steps in TypeTopology, and at least 34 occurrences of reflexive equivalence steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Please add yourself the first time you contribute.
 * Ayberk Tosun
 * Brendan Hart
 * Bruno Paiva
+* Carlo Angiuli
 * Chuangjie Xu
 * Cory Knapp
 * Ettore Aldrovandi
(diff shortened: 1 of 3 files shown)


martinescardo:
There is a folder called `gist` in `TypeTopology`.

PR #399, opened 2026-01-04
Reflexive graphs

IanRay11:
I commence the (rather large) PR pertaining to reflexive graphs. I will only submit one file at a time. Be aware that this formalization follows Jon Sterling's paper "Reflexive graph lenses" and I have initially adopted his terminology. But I am willing to adjust this to be more descriptive and match the type topology style (this isn't relevant for this PR but will be later).

### DIFF
--- a/source/ReflexiveGraphs/index.lagda
+++ b/source/ReflexiveGraphs/index.lagda
@@ -4,7 +4,7 @@ The structure identity principle (SIP), coined by Peter Aczel, allows an
 treatment of identificiation in Univalent Foundations that, with much care,
 escapes "transport hell". Many have formulated there own terminology and
 approach to SIP (including Egbert Rijke in "Introduction to Homotopy Type
-Theory" see doi:10.1017/9781108933568, arXiv:2212.11082; Martin Escardo see
+Theory" see doi:10.1017/9781108933568, arXiv:2212.11082; Martin Escardo in
 files: UF/SIP, UF/Yoneda and UF/SigmaIdentity; as well as many others!) In
 recent times, some have considered 'reflexive graphs' as a more modular
 approach to SIP (see "Using Displayed Univalent Graphs to Formalize Higher


ayberkt:
@IanRay11 Is this ready to be reviewed? Otherwise, can I mark it as draft?

source/ReflexiveGraphs/ReflexiveGraphs.lagda
```diff
@@ -0,0 +1,78 @@
...
+The structure identity principle (SIP), coined by Peter Aczel, allows an
+treatment of identificiation in Univalent Foundations that, with much care,
+escapes "transport hell". Many have formulated there own terminology and
+approach to SIP (including Egbert Rijke in "Introduction to Homotopy Type
+Theory"; Martin Escardo see files: StructureIdentityPrinciple, Yoneda and
+SigmaIdentity; as well as many others!) In recent times, some have
+considered 'reflexive graphs' as a more systematic approach to SIP (see
+"Using Displayed Univalent Graphs to Formalize Higher Groups in Univalent
+Foundations" by Johannes Schipp von Branitz and Ulrik Buchholtz; and
+"Reflexive graph lenses in univalent foundations" by Jonathan Sterling).
```
tomdjong: Maybe you could include DOIs for these documents.

source/ReflexiveGraphs/ReflexiveGraphs.lagda
```diff
@@ -0,0 +1,78 @@
...
+
+We will develop a portion of the theory of reflexive graphs here while
+primarily following Jonathon Sterling's treatment from the aformentioned
```
tomdjong: Could you comment on where you deviate from Jon's treatment?

source/ReflexiveGraphs/ReflexiveGraphs.lagda
```diff
@@ -0,0 +1,78 @@
...
+ refl-graph : (𝓤 ⊔ 𝓥)⁺ ̇
+ refl-graph = Σ A ꞉ 𝓤 ̇ , Σ R ꞉ (A → A → 𝓥 ̇) , ((x : A) → R x x)
+
+\end{code}
+
+We give some boiler plate/syntax
+
+\begin{code}
+
+⊰_⊱ : refl-graph 𝓤 𝓥 → 𝓤 ̇
```
tomdjong: Maybe use `⟨_⟩` instead of `⊰_⊱` in line with most of TypeTopology so far.

source/ReflexiveGraphs/ReflexiveGraphs.lagda
```diff
@@ -0,0 +1,78 @@
...
+edge-rel (_ , R , _) = R
+
+syntax edge-rel 𝓐 x y = x ≈⟨ 𝓐 ⟩ y
+
+𝓻 : (𝓐 : refl-graph 𝓤 𝓥) → (x : ⊰ 𝓐 ⊱) → x ≈⟨ 𝓐 ⟩ x
+𝓻 (_ , _ , r) x = r x
+
+\end{code}
+
+We define a homomorphism of reflexive graphs as a sigma and record type.
```
tomdjong: a homomorphism -> homomorphisms

source/ReflexiveGraphs/ReflexiveGraphs.lagda
```diff
@@ -0,0 +1,78 @@
...
+\begin{code}
+
+module _ (𝓤 𝓥 : Universe) where
+
+ refl-graph : (𝓤 ⊔ 𝓥)⁺ ̇
+ refl-graph = Σ A ꞉ 𝓤 ̇ , Σ R ꞉ (A → A → 𝓥 ̇) , ((x : A) → R x x)
+
+\end{code}
+
+We give some boiler plate/syntax
```
tomdjong: Missing full stop.

source/ReflexiveGraphs/ReflexiveGraphs.lagda
```diff
@@ -0,0 +1,78 @@
...
+
+refl-graph-hom : (𝓐 : refl-graph 𝓤 𝓥) (𝓐' : refl-graph 𝓤' 𝓥')
+               → 𝓤 ⊔ 𝓥 ⊔ 𝓤' ⊔ 𝓥' ̇
+refl-graph-hom 𝓐 𝓐'
+ = Σ F ꞉ (⊰ 𝓐 ⊱ → ⊰ 𝓐' ⊱) ,
+    Σ F' ꞉ ((x y : ⊰ 𝓐 ⊱) → x ≈⟨ 𝓐 ⟩ y → F x ≈⟨ 𝓐' ⟩ F y) ,
+     ((x : ⊰ 𝓐 ⊱) → F' x x (𝓻 𝓐 x) ＝ 𝓻 𝓐' (F x))
+
+record refl-graph-hom-record
+ (𝓐 : refl-graph 𝓤 𝓥) (𝓐' : refl-graph 𝓤' 𝓥') : 𝓤ω where
```
tomdjong: The `𝓤ω` should be replaced by `𝓤 ⊔ 𝓥 ⊔ 𝓤' ⊔ 𝓥' ̇`.

source/ReflexiveGraphs/ReflexiveGraphs.lagda
tomdjong: If it's 'just' a formalization of Jon's paper, then it would be better to just explicitly say so in my opinion.

IanRay11: See global comment.

tomdjong: But this particular file follows Jon's paper, right? So it would be good if it said so. A good place for your global comment would be the index file for this folder, I think.

jonsterling: Jonathon => Jonathan ;-)

IanRay11: I'm sorry!

IanRay11: Correct. And that is a good point. I think the index file would be a better place. Initially I thought here because when someone looks at the formalization they should start here, but on second thought they may not know that!

jonsterling: not to worry haha ;-)

source/ReflexiveGraphs/ReflexiveGraphs.lagda
```diff
@@ -0,0 +1,68 @@
...
+⟨_⟩ : refl-graph 𝓤 𝓥 → 𝓤 ̇
+⟨ (A , _) ⟩ = A
+
+edge-rel : (𝓐 : refl-graph 𝓤 𝓥) → ⟨ 𝓐 ⟩ → ⟨ 𝓐 ⟩ → 𝓥 ̇
+edge-rel (_ , R , _) = R
+
+syntax edge-rel 𝓐 x y = x ≈⟨ 𝓐 ⟩ y
+
+𝓻 : (𝓐 : refl-graph 𝓤 𝓥) → (x : ⟨ 𝓐 ⟩) → x ≈⟨ 𝓐 ⟩ x
+𝓻 (_ , _ , r) x = r x
```
jonsterling: Maybe it is easier to type `rx` than `𝓻`? But I don't feel strongly, especially if the 𝓻 is meant to conform with existing TypeTopology practice.

source/ReflexiveGraphs/ReflexiveGraphs.lagda
```diff
@@ -0,0 +1,68 @@
...
+ = Σ F ꞉ (⟨ 𝓐 ⟩ → ⟨ 𝓐' ⟩) ,
+    Σ F' ꞉ ((x y : ⟨ 𝓐 ⟩) → x ≈⟨ 𝓐 ⟩ y → F x ≈⟨ 𝓐' ⟩ F y) ,
+     ((x : ⟨ 𝓐 ⟩) → F' x x (𝓻 𝓐 x) ＝ 𝓻 𝓐' (F x))
+
+record refl-graph-hom-record
+ (𝓐 : refl-graph 𝓤 𝓥) (𝓐' : refl-graph 𝓤' 𝓥') : 𝓤 ⊔ 𝓥 ⊔ 𝓤' ⊔ 𝓥' ̇ where
+ field
+  func : ⟨ 𝓐 ⟩ → ⟨ 𝓐' ⟩
+  act : (x y : ⟨ 𝓐 ⟩) → x ≈⟨ 𝓐 ⟩ y → func x ≈⟨ 𝓐' ⟩ func y
+  pres-ref : (x : ⟨ 𝓐 ⟩) → act x x (𝓻 𝓐 x) ＝ 𝓻 𝓐' (func x)
```
jonsterling: Could this be renamed to `pres-XXX` where `XXX` is whatever the reflexivity datum is called?

source/ReflexiveGraphs/ReflexiveGraphs.lagda
IanRay11: Just to check, what is rx short for?

jonsterling: It’s short for reflexivity.

IanRay11: I thought so. I just wanted to make sure it was not like the x that we are reflexive at or something. 

I am in favor of an easier to type symbol as opposed to *cool r*.

tomdjong: I wonder if `rx x` might look confusing. Bit long, but what about `gr-refl`?

source/ReflexiveGraphs/index.lagda
```diff
@@ -0,0 +1,28 @@
+Ian Ray.
+
+The structure identity principle (SIP), coined by Peter Aczel, allows an
+treatment of identificiation in Univalent Foundations that, with much care,
+escapes "transport hell". Many have formulated there own terminology and
+approach to SIP (including Egbert Rijke in "Introduction to Homotopy Type
```
tomdjong, line 6: doi:10.1017/9781108933568, arXiv:2212.11082

Since links are not clickable anyway, I'm just suggesting the give the identifiers here (to the published and the arXiv version).

source/ReflexiveGraphs/index.lagda
```diff
@@ -0,0 +1,28 @@
...
+Theory"; Martin Escardo see files: StructureIdentityPrinciple, Yoneda and
+SigmaIdentity; as well as many others!) In recent times, some have
+considered 'reflexive graphs' as a more systematic approach to SIP (see
```
tomdjong: Out of curiosity, what do you mean by *more systematic*?

source/ReflexiveGraphs/ReflexiveGraphs.lagda
tomdjong: Maybe you can write a comment explaining why `act` is called `act` (presumably in analogy with *action on paths*?) rather than `pres-≈` for instance?

source/ReflexiveGraphs/index.lagda
IanRay11: Perhaps I should say "modular". Using displayed reflexive graphs we can break the process of characterizing sigma types (which by the identity system approach is already quite well understood and systematic) into an arguably even more principled and predictable set of steps. Additionally, the goals of each of the steps are typically immediate from the closure properties enjoyed by univalent reflexive graphs.

martinescardo: Also, the correct file is `UF.SIP`. The other one is obsolete (and more convoluted than necessary).

source/ReflexiveGraphs/index.lagda
martinescardo: (But they hopefully will become clickable in the future.)

source/ReflexiveGraphs/ReflexiveGraphs.lagda
martinescardo: Yes, we prefer longer, clear names. I would even expand `gr` in `gr-refl` to something more meaningful.

martinescardo: Oh, what about `≈-refl`? We use a similar notation for the identity equivalence.

source/ReflexiveGraphs/index.lagda
```diff
@@ -0,0 +1,28 @@
...
+of structures in a principled way using (displayed) (univalent) reflexive
+graphs.
+
+\begin{code}
+
+{-# OPTIONS --safe --without-K #-}
+
+module ReflexiveGraphs.index where
+
+import ReflexiveGraphs.ReflexiveGraphs
```
martinescardo: In general, we avoid things like `Groups.Group` etc., and we write `Groups.Type` or `Groups.Definition`. I think the same naming convention should be applied here.

martinescardo: So I suggest renaming this file to `Type.lagda` or `Definition.lagda`. I don't mind which, but most of the time I use the former, although I use the latter when it makes more sense.

source/ReflexiveGraphs/ReflexiveGraphs.lagda
tomdjong: I like Martín's latest suggestion best.

martinescardo (approved):
@tomdjong , I am giving my final approval for this. Please merge if you agree.

source/ReflexiveGraphs/Type.lagda
```diff
@@ -0,0 +1,68 @@
...
+\end{code}
+
+A reflexive graph consists of a type, a binary type valued relation and a
+reflexivity datum.
+
+\begin{code}
+
+module _ (𝓤 𝓥 : Universe) where
+
+ refl-graph : (𝓤 ⊔ 𝓥)⁺ ̇
```
martinescardo: Oh, one final thought. I don't mind the current chosen notation `refl-graph`.

But we apply the unstated convention that we capitalize names of types of mathematical structures, such as `Group`. I don't feel strongly about this, but I would prefer the capitalized version.

IanRay11: Ahh yes. I agree.

martinescardo: When you are done addressing this, I'll merge.

tomdjong:
I can do a final check tomorrow. But wasn't there something about `UF.SIP` still?

source/ReflexiveGraphs/index.lagda
IanRay11: @tomdjong I think I corrected this.

source/ReflexiveGraphs/index.lagda
```diff
@@ -0,0 +1,27 @@
+Ian Ray.
+
+The structure identity principle (SIP), coined by Peter Aczel, allows an
+treatment of identificiation in Univalent Foundations that, with much care,
+escapes "transport hell". Many have formulated there own terminology and
+approach to SIP (including Egbert Rijke in "Introduction to Homotopy Type
+Theory" see doi:10.1017/9781108933568, arXiv:2212.11082; Martin Escardo see
+files: SIP, Yoneda and SigmaIdentity; as well as many others!) In recent
```
tomdjong: Since these files do not live in the same folder, namely `ReflexiveGraphs`, but in `UF` instead, could you add `UF.`  or `UF/` as a prefix please?

source/ReflexiveGraphs/index.lagda
```diff
@@ -0,0 +1,27 @@
+Ian Ray.
```
tomdjong: Could you add a date here as well please? Or perhaps a range of dates, e.g. August 2025–February 2026?

source/ReflexiveGraphs/Type.lagda
```diff
@@ -0,0 +1,68 @@
+Ian Ray. 28th August 2025.
+
```
tomdjong, line 2: Maybe add "Minor changes and merged into TypeTopology in February 2026"?

source/ReflexiveGraphs/index.lagda
IanRay11: Good point.

IanRay11:
@ayberkt I'm not sure why the check failed. It shouldn't be the files (I only changed a comments since the last passed check).

ayberkt:
@IanRay11 Let’s see if the checks pass for commit [`7b623e6`](https://github.com/martinescardo/TypeTopology/pull/399/commits/7b623e6802062cab99706bcec79713ac8253d595). I’ll look into it if it doesn’t.

ayberkt:
@IanRay11 This seems to be the result of a large-scale [incident](https://www.githubstatus.com/?todayis=2026-02-02) at GitHub, being investigated at the moment. There is nothing I can do. Let’s just wait until it’s resolved.

PR #400, opened 2026-01-08
Index file for our injectives paper

tomdjong:


source/DomainTheory/Examples/ExtendedPartialDedekindReals.lagda
```diff
@@ -0,0 +1,196 @@
...
+  U-is-upper : (x : 𝓡∞) → is-upper (U x)
+  U-is-upper x = pr₁ (pr₁ (pr₂ (pr₂ x)))
+
+  U-is-lower-open : (x : 𝓡∞) → is-lower-open (U x)
+  U-is-lower-open x = pr₂ (pr₁ (pr₂ (pr₂ x)))
+
+  orderedness : (x : 𝓡∞) → are-ordered (L x) (U x)
+  orderedness x = pr₂ (pr₂ (pr₂ x))
+
+  -- TODO(?). Move this and the equivalence to Various.Dedekind?
```
martinescardo: I think it would make sense to do this.

tomdjong: Done.

source/Taboos/BasicDiscontinuity.lagda
```diff
@@ -183,6 +183,30 @@ basic-discontinuity-taboo' f (f₀ , f₁) = VI
           (λ (p : ι 0 ＝ f u) → inr (I u (p ⁻¹)))
           (λ (ν : ι 0 ≠ f u) → inl (II u (≠-sym ν)))
 
+WLPO-is-discontinuous' : WLPO
```
martinescardo, line 186: Isn't this somewhere already? I'll check when I have time.

martinescardo: Maybe not exactly this formulation.

tomdjong: The unprimed version is there already (and I made use of that). The primed version isn't, I think. 
`grep -R "basic-discontinuity'"` only returns the definition and what I just added.

PR #403, opened 2026-02-03
Basic Category Theory

awsloth:


tomdjong:
I can take a look, but it might take a while, given the number of files (even only counting the new ones).

PR #406, opened 2026-02-04
Univalent refl graph

IanRay11:
Again, we are following Jon's terminology. 

Some things to note: I tried to address things that naturally carry over from the previous PR (capitalize the type of univalent refl graphs, use more standard bracket for carrier projection, rename reflexivity component more in line with TypeTopology convention, etc.) but it is possible I missed somethings. If so I apologize for making you readdress something redundant.

### DIFF
--- a/source/ReflexiveGraphs/index.lagda
+++ b/source/ReflexiveGraphs/index.lagda
@@ -24,6 +24,6 @@ that outline the reflexive graph approach to SIP.
 module ReflexiveGraphs.index where
 
 import ReflexiveGraphs.Type
-import ReflexiveGraphs.UnivalentReflexiveGraphs 
+import ReflexiveGraphs.Univalent
 
 \end{code}


tomdjong:
@IanRay11 Thanks for this! I'll write more later but for now `.././imports.sh ReflexiveGraphs/UnivalentReflexiveGraphs.lagda` correctly reports
```
Importing Notation.General was not necessary
Importing UF.Base was not necessary
Importing UF.FunExt was not necessary
```

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+
+Minor changes and merged into TypeToplogy in February 2026.
+
+We give some properties about fans (terminology borrowed from Sterling),
+which are analogous to singletons up to the edge relation. Then we provide
+some equivalent characterizations of univalent reflexive graphs. It is worth
+noting that, although Sterling makes no choice for the defintion in his paper,
+we are required to. There is good reason to go with the 'propositional fans'
+definition as it simplifies many proofs later on, but all of the definitions
+are useful (I would be willing to consider making a different choice).
```
tomdjong: > (I would be willing to consider making a different choice).

I guess this is supposed to be removed post-review?

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+
+We provide some equivalent descriptions of univalent reflexive graphs (see
+Sterling, Ulrik, etc.)
```
tomdjong: Ulrik → Buchholtz (in line with Sterling). 
I would also include a line explaining that you are referring to works mentioned in `index.lagda`. On their own, these names are not helpful.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+{-# OPTIONS --safe --without-K #-}
+
+module ReflexiveGraphs.UnivalentReflexiveGraphs where
+
+open import MLTT.Spartan
+open import Notation.General
+open import UF.Base
+open import UF.Equiv
+open import UF.EquivalenceExamples
+open import UF.FunExt
```
tomdjong: See my comment about the redundant imports.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+      → ⟨ 𝓐 ⟩
+      → 𝓤 ⊔ 𝓥 ̇ 
+cofan 𝓐 x = Σ y ꞉ ⟨ 𝓐 ⟩ , y ≈⟨ 𝓐 ⟩ x
+
+prop-fan-to-cofan : (𝓐 : Refl-Graph 𝓤 𝓥)
+                  → ((x : ⟨ 𝓐 ⟩) → is-prop (fan 𝓐 x))
+                  → ((x : ⟨ 𝓐 ⟩) → is-prop (cofan 𝓐 x))
+prop-fan-to-cofan {𝓤} {𝓥} 𝓐 fan-prop = I ∼-refl
+ where
+  I = ((x : ⟨ 𝓐 ⟩) → is-prop (cofan 𝓐 x))
```
tomdjong: I'm sorry, but I'm struggling to read this proof. Would it not be easier to assume `x : ⟨ 𝓐 ⟩`, two elements of `cofan 𝓐 x` and prove that they're equal using `fan-prop`?

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+           → x ≈⟨ 𝓐 ⟩ y
+id-to-edge 𝓐 {x} {x} refl = ≈-refl 𝓐 x
+
+\end{code}
+
+If each fan is propositional then id-to-edge has a section and retraction.
+
+\begin{code}
+
+helper-edge-to-id : {𝓐 : Refl-Graph 𝓤 𝓥}
```
tomdjong: Maybe make this one `private`?

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+\begin{code}
+
+id-to-edge : (𝓐 : Refl-Graph 𝓤 𝓥) {x y : ⟨ 𝓐 ⟩}
+           → x ＝ y
+           → x ≈⟨ 𝓐 ⟩ y
+id-to-edge 𝓐 {x} {x} refl = ≈-refl 𝓐 x
+
+\end{code}
+
+If each fan is propositional then id-to-edge has a section and retraction.
```
tomdjong: So it's an equivalence. Just say that instead?

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+open import UF.Base
+open import UF.Equiv
+open import UF.EquivalenceExamples
+open import UF.FunExt
+open import UF.PropIndexedPiSigma
+open import UF.Retracts
+open import UF.Subsingletons
+open import UF.Subsingletons-Properties
+open import ReflexiveGraphs.Type
+
```
tomdjong: You can place nearly everything in an anonymous module with `(𝓐 : Refl-Graph 𝓤 𝓥)` as a parameter to avoid having to introduce it all the time.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+\begin{code}
+
+helper-edge-to-id : {𝓐 : Refl-Graph 𝓤 𝓥}
+                  → (x y : ⟨ 𝓐 ⟩)
+                  → (p : x ≈⟨ 𝓐 ⟩ y)
+                  → (x , ≈-refl 𝓐 x) ＝ (y , p)
+                  → x ＝ y
+helper-edge-to-id {_} {_} {𝓐} x .x .(≈-refl 𝓐 x) refl = refl
+
```
tomdjong: Consider making `((x : ⟨ 𝓐 ⟩) → is-prop (fan 𝓐 x))` an anonymous module parameter.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+We now define univalent reflexive graphs in terms of propositional fans, but
+one could use any of the equivalent characterizations.
+
+\begin{code}
+
+is-univalent-refl-graph : (𝓐 : Refl-Graph 𝓤 𝓥) → 𝓤 ⊔ 𝓥 ̇ 
+is-univalent-refl-graph 𝓐 = (x : ⟨ 𝓐 ⟩) → is-prop (fan 𝓐 x)
+
+Univalent-Refl-Graph : (𝓤 𝓥 : Universe) → (𝓤 ⁺) ⊔ (𝓥 ⁺) ̇
```
tomdjong: Replace `(𝓤 ⁺) ⊔ (𝓥 ⁺) ̇` by ` (𝓤 ⊔ 𝓥) ⁺ ̇`.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+≈-reflᵤ : (𝓐 : Univalent-Refl-Graph 𝓤 𝓥) → (x : ⟨ 𝓐 ⟩ᵤ) → x ≈ᵤ⟨ 𝓐 ⟩ x
+≈-reflᵤ (𝓐 , _) x = ≈-refl 𝓐 x
+
+underlying-refl-graph : (𝓐 : Univalent-Refl-Graph 𝓤 𝓥)
+                      → Refl-Graph 𝓤 𝓥
+underlying-refl-graph (𝓐 , _) = 𝓐
+
+syntax underlying-refl-graph 𝓐 = 𝓐 /ᵤ 
+
+underlying-univalence : (𝓐 : Univalent-Refl-Graph 𝓤 𝓥)
```
tomdjong: Maybe `underlying-refl-graph-is-univalent` instead of `underlying-univalence?

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+                      → is-univalent-refl-graph (𝓐 /ᵤ)
+underlying-univalence (𝓐 , is-ua) = is-ua
+
+id-equiv-edge : (𝓐 : Univalent-Refl-Graph 𝓤 𝓥)
```
tomdjong: What is called `id` now, was called `paths` earlier. Please make a consistent choice.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+id-equiv-edge 𝓐 x y
+ = (id-to-edge (𝓐 /ᵤ)
+   , prop-fans-implies-id-to-edge-equiv (underlying-univalence 𝓐) x y)
+
+edge-to-id : (𝓐 : Univalent-Refl-Graph 𝓤 𝓥) {x y : ⟨ 𝓐 ⟩ᵤ}
+           → x ≈ᵤ⟨ 𝓐 ⟩ y
+           → x ＝ y
+edge-to-id 𝓐 {x} {y} = ⌜ id-equiv-edge 𝓐 x y ⌝⁻¹
+
+edge-to-id-comp : (𝓐 : Univalent-Refl-Graph 𝓤 𝓥) {x : ⟨ 𝓐 ⟩ᵤ}
```
tomdjong: `edge-to-id-preserves-refl` instead?

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,314 @@
...
+
+\end{code}
+
+We consider the notion of edge induction and show univalence implies it.
+
+TODO: show they are also equivalent.
+
+\begin{code}
+
+edge-induction : (𝓣 : Universe) (𝓐 : Refl-Graph 𝓤 𝓥) → 𝓤 ⊔ 𝓥 ⊔ (𝓣 ⁺) ̇
```
tomdjong: Maybe this range over all universes instead (like in a proper induction principle). Or is it going to be interesting to track for which `𝓣` you need it?

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
IanRay11: fan-prop only allows us to identify fans no cofans. That's why much of the proof is massaging the data around so that a cofan turns into a fan, etc.

tomdjong: Yes, I understand that, but what I said should still apply to your proof. (I did not mean to imply that the proof was trivial.)

IanRay11: I may have proved it that way originally. I think I ended up with the current proof because it was more faithful to the paper proof that Jon gives.

jonsterling: Feel free to use whatever proof is better ;-)

IanRay11: Actually, I'm not sure I understand the proposed proof. How am I to use fan-prop to show to terms of cofan are equal?

tomdjong: Let me try to explain myself better. Please let me know if I'm either misunderstanding something, or if it's still unclear what I'm asking about.
You've got a statement of the form: `(x : X → A x) → (x : X → B x)`. (With `A` saying that the type of fans at `x` is a subsingleton and `B` is saying the same thing but about cofans this time.) Ordinarily, you would assume `h : (x : X → A x)` and `x : X`, and then prove `B x` somehow. Do we agree so far?
In the current proof you never really introduce `x : X` and constantly work with type `x : X → B x` rather than `B x` for a fixed but arbitrary `x`. Am I seeing that correctly? If so, why do you write the proof this way? The issue I have with the current proof is that I'm finding it hard to read (sorry!).
(Since, in this particular case, we are talking about proving that something is a subsingleton, you can further bring two elements of the relevant type into scope.)

IanRay11: I do see your point. The benefit to not fixing x here is it allows us to turn a something of type cofan x to something of type cofan y. In particular, for any x if (y,s) : cofan x then (x,s) : fan y. Then we contract this part away by the assumption (and do something similar with the remaining (y',t) pair.) At one point the proof was instead a string of equivalences but that requires function extensionality so we opt to just observe the underlying maps of the equivalence as a string of reductions.

Also, originally I did prove it in the way you describe. Where I brought every element into the scope and tried to show the resulting pair was equal but it required a weird and unmotivated transport. You can look at the commit history and compare (maybe this approach could also be improved: see line 57 of the univalentreflexivegraph file https://github.com/IanRay11/TypeTopology/commit/7690d1f0a3650a1876165994479ed6006853ed4c)

tomdjong: Thanks, I see, it's a nice proof. Could you please
(1) write a brief comment explaining that there is a proof by equivalences which you avoid because that relies on function extensionality,
(2) format the proof like I partially did below?
```
   I =
    ((x : ⟨ 𝓐 ⟩) → is-prop (cofan x))                                  ←⟨ id ⟩
    ((x : ⟨ 𝓐 ⟩) → ((y , s) (y' , t) : cofan x) → (y , s) ＝ (y' , t)) ←⟨ II ⟩
    ((x y : ⟨ 𝓐 ⟩) (s : y ≈⟨ 𝓐 ⟩ x) (y' : ⟨ 𝓐 ⟩) (t : y' ≈⟨ 𝓐 ⟩ x)
      → (y , s) ＝ (y' , t))                                           ←⟨ III ⟩
    [...]
    where
     II  = λ f x (y , s) (y' , t) → f x y s y' t
     III = λ f x y → f y x
     [...]
```
Hopefully you'll agree that this is a little easier on the eye.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,304 @@
...
+
+ cofan : ⟨ 𝓐 ⟩
+       → 𝓤 ⊔ 𝓥 ̇ 
+ cofan x = Σ y ꞉ ⟨ 𝓐 ⟩ , y ≈⟨ 𝓐 ⟩ x
+
+\end{code}
+
+We show that propositional fans imply propositional cofans and vice versa.
+The crux of the proofs is to observe that if (y , s) : cofan x for some x then
+(x , s) : fan y, the rest amounts to shuffling data with in sigma types.
```
tomdjong: _with in_ → _within_
But since there is no explicit mention of sigma types it might be less confusing to just say _shuffling data_.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,304 @@
...
+
+It is worth noting that the proofs that follow were originally discovered as
+a string of equivalences, but to witness the equivalence requires function
+extensionality. The underlying maps of the equivalences are sufficient for the
+proof to go through and thus we are able to avoid unnecessary appeals to
```
tomdjong: (Minor) _unncessary_ is redundant here; if the appeals were necessary you couldn't avoid them.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,304 @@
...
+ id-to-edge {x} {x} refl = ≈-refl 𝓐 x
+
+\end{code}
+
+If each fan is propositional then id-to-edge is an equivalence.
+
+\begin{code}
+
+ private
+  helper-edge-to-id : (x y : ⟨ 𝓐 ⟩)
```
tomdjong: Maybe the argument `x` and `y` should be implicit, like in `id-to-edge`?

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,304 @@
...
+                    → (p : x ≈⟨ 𝓐 ⟩ y)
+                    → (x , ≈-refl 𝓐 x) ＝ (y , p)
+                    → x ＝ y
+  helper-edge-to-id x .x .(≈-refl 𝓐 x) refl = refl
+
+ module _ (prop-fans : ((x : ⟨ 𝓐 ⟩) → is-prop (fan x))) where
```
tomdjong: It would make sense to include `(x y : ⟨ 𝓐 ⟩)` as well as that assumption is common in what follows.
Please format it like this
```
module _
        (one : ...)
        (two : ...)
       where
```

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,304 @@
...
+    ((x : ⟨ 𝓐 ⟩) → ((y , s) (y' , t) : cofan x) → (y , s) ＝ (y' , t)) ←⟨ II ⟩ 
+    ((x y : ⟨ 𝓐 ⟩) (s : y ≈⟨ 𝓐 ⟩ x) (y' : ⟨ 𝓐 ⟩) (t : y' ≈⟨ 𝓐 ⟩ x)
+      → (y , s) ＝ (y' , t))                                           ←⟨ III ⟩
+    ((y x : ⟨ 𝓐 ⟩) (s : y ≈⟨ 𝓐 ⟩ x) (y' : ⟨ 𝓐 ⟩) (t : y' ≈⟨ 𝓐 ⟩ x)
+      → (y , s) ＝ (y' , t))                                           ←⟨ IV ⟩
+    ((y : ⟨ 𝓐 ⟩) ((x , s) : fan y) (y' : ⟨ 𝓐 ⟩) (t : y' ≈⟨ 𝓐 ⟩ x)
+      → (y , s) ＝ (y' , t))                                           ←⟨ V ⟩
+    ((y y' : ⟨ 𝓐 ⟩) (t : y' ≈⟨ 𝓐 ⟩ y) → (y , ≈-refl 𝓐 y) ＝ (y' , t))  ←⟨ VI ⟩
+    ((y' y : ⟨ 𝓐 ⟩) (t : y' ≈⟨ 𝓐 ⟩ y) → (y , ≈-refl 𝓐 y) ＝ (y' , t))  ←⟨ VII ⟩
+    ((y' : ⟨ 𝓐 ⟩) ((y , t) : fan y') → (y , ≈-refl 𝓐 y) ＝ (y' , t))   ←⟨ VIII ⟩
```
tomdjong: Please align all `←⟨` with the `▢` (at column 73, according to my emacs).
Same for `prop-cofan-to-fan`.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
tomdjong: Otherwise, thanks very much for the comment and explanation!

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,304 @@
...
+    II p = I p (prop-fans x (x , ≈-refl 𝓐 x) (y , p))
+
+  edges-are-retracts-of-id : (x y : ⟨ 𝓐 ⟩)
+                           → retract (x ≈⟨ 𝓐 ⟩ y) of (x ＝ y)
+  edges-are-retracts-of-id x y
+   = (id-to-edge , prop-fans-gives-section x y)
+
+\end{code}
+
+Now we show that id-to-edge is an equiv iff all fans are propositional.
```
tomdjong: This now almost says the same as the comment on line 146 (although this mentions the converse). Maybe one comment is all you need?

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,304 @@
...
+edge-to-id-preserves-refl 𝓐 {x}
+ = inverses-are-retractions (id-to-edge (𝓐 /ᵤ))
+    (prop-fans-implies-id-to-edge-equiv (underlying-refl-graph 𝓐)
+     (underlying-refl-graph-is-univalent 𝓐) x x) refl
+
+\end{code}
+
+We show univalence implies edge induction.
+
+TODO: show they are also equivalent.
```
tomdjong: For now, it's OK, but when you add the converse, then you should make a definition of `edge-induction`, so that you can write `is-univalent-refl-graph 𝓐 → edge-induction 𝓐` and its converse.

tomdjong (commented):
Thanks for the changes @IanRay11! As far as I'm concerned it should be ready for merging after these minor comments are addressed.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
IanRay11: Maybe mine is off. I'll try to sort this out.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
IanRay11: There are actually technical reasons why I didn't want x and y in the module. I can rework the proofs if you feel strongly about this.

IanRay11: In the where column of prop-fans-gives-retraction I need to use prop-fans-edge-to-id on a non-fixed x and y. I spent no time thinking about how to adjust the proofs, but I can do this if you would like.

tomdjong: You can still do the required pattern match without reworking, i.e. the following just works
```
  prop-fans-gives-retraction : has-retraction id-to-edge
  prop-fans-gives-retraction = (prop-fans-edge-to-id , II)
   where
    I : (x : ⟨ 𝓐 ⟩) → prop-fans x (x , ≈-refl 𝓐 x) (x , ≈-refl 𝓐 x) ＝ refl
    I x = props-are-sets (prop-fans x)
           (prop-fans x (x , ≈-refl 𝓐 x) (x , ≈-refl 𝓐 x)) refl
    II : (p : x ＝ y) → prop-fans-edge-to-id (id-to-edge p) ＝ p
    II refl = ap (helper-edge-to-id (≈-refl 𝓐 x)) (I x)
```
as does this where I fixed `x` in `I`:
```
  prop-fans-gives-retraction : has-retraction id-to-edge
  prop-fans-gives-retraction = (prop-fans-edge-to-id , II)
   where
    I : prop-fans x (x , ≈-refl 𝓐 x) (x , ≈-refl 𝓐 x) ＝ refl
    I = props-are-sets (prop-fans x) 
         (prop-fans x (x , ≈-refl 𝓐 x) (x , ≈-refl 𝓐 x))
         refl
    II : (p : x ＝ y) → prop-fans-edge-to-id (id-to-edge p) ＝ p
    II refl = ap (helper-edge-to-id (≈-refl 𝓐 x)) I
```

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
tomdjong: Right now, half are at 71, half are at 70, and the square is at 72 :sweat_smile:

tomdjong: If you keep having trouble with this, I can just edit it after it's merged.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
IanRay11: Thank you for this. Again I'm sorry for not giving it more thought I was a bit preoccupied with something else. I have implemented your suggestions.

tomdjong:
@IanRay11 I made some minor modifications so that I don't have to do it post-merging, so you should pull.

source/ReflexiveGraphs/index.lagda
```diff
@@ -24,5 +24,6 @@ that outline the reflexive graph approach to SIP.
 module ReflexiveGraphs.index where
 
 import ReflexiveGraphs.Type
+import ReflexiveGraphs.UnivalentReflexiveGraphs
```
martinescardo: I feel that this module should be called just `Univalent`, so that the import reads
```
import ReflexiveGraphs.Univalent
```

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,298 @@
+Ian Ray. 2nd September 2025.
+
+Minor changes and merged into TypeToplogy in February 2026.
```
martinescardo: Typo: "TypeToplogy"

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,298 @@
...
+
+We give some properties about fans (terminology borrowed from Sterling),
+which are analogous to singletons up to the edge relation. Then we provide
+some equivalent characterizations of univalent reflexive graphs. It is worth
+noting that, although Sterling makes no choice for the defintion in his paper,
+we are required to. There is good reason to go with the 'propositional fans'
+definition as it simplifies many proofs later on, but all the definitions
+are useful.
+
+We provide some equivalent descriptions of univalent reflexive graphs (see
```
martinescardo: I feel that this should be the first paragraph of the description of this module.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
```diff
@@ -0,0 +1,298 @@
...
+open import UF.PropIndexedPiSigma
+open import UF.Retracts
+open import UF.Subsingletons
+open import UF.Subsingletons-Properties
+open import ReflexiveGraphs.Type
+
+module _ (𝓐 : Refl-Graph 𝓤 𝓥) where
+
+ fan : ⟨ 𝓐 ⟩
+     → 𝓤 ⊔ 𝓥 ̇
```
martinescardo: Probably no need to break the line here (and in similar situations below), but I am OK keeping this as it is if you prefer.

source/ReflexiveGraphs/UnivalentReflexiveGraphs.lagda
martinescardo: (The one starting in line 13.)

PR #432, opened 2026-04-06
Univalent families redo

IanRay11:
Univalent families


IanRay11:
I will check for unused imports then request review.

IanRay11:
@tomdjong

source/ReflexiveGraphs/UnivalentFamilies.lagda
```diff
@@ -0,0 +1,79 @@
...
+
+is-univalent-family : Σ A ꞉ 𝓤 ̇ , (A → 𝓣 ̇)
+                    → 𝓤 ⊔ 𝓣 ̇
+is-univalent-family (A , B) = is-univalent-refl-graph (refl-graph-image A B)
+
+\end{code}
+
+This definition is intentionally general and offers a robust theoretical
+framework for working with a universe of things. That is, if one was working
+directly with the codes of a tarski style universe presentation than a
```
tomdjong: _**T**arski_; _then_

source/ReflexiveGraphs/UnivalentFamilies.lagda
```diff
@@ -0,0 +1,79 @@
...
+univalent family would be a pair (𝓤 , El) consiting of the universe type and
+the decoding map El : 𝓤 → Type. In Agda the user may operate as if universes
+where presented à la Russel and thus our pair of interest is simply (𝓤 , id).
+But there are other examples of univalent families of interest such as
+(Prop, pr₁), (Set, pr₁), (ℕ , finset), etc.
```
tomdjong: Presumably prime examples would be mathematical structures, e.g. for magmas: the universe together with the map `X ↦ (X → X → X)`, cf. Martín's SIP.

source/ReflexiveGraphs/UnivalentFamilies.lagda
```diff
@@ -0,0 +1,79 @@
...
+
+We can give some equivalent characterizations of univalent family of types.
```
tomdjong, line 51: You don't actually establish an equivalence, only one direction.

source/ReflexiveGraphs/UnivalentFamilies.lagda
```diff
@@ -0,0 +1,79 @@
+Ian Ray. 28th August 2025.
+
+Minor changes and merged into TypeToplogy in April 2026.
+
+We define the notion of univalent family in terms of a reflexive graph image.
```
tomdjong, line 5: Is the terminology "reflexive graph image" due to Jon as well?

tomdjong (commented):
Thanks for making small PRs, Ian!

source/ReflexiveGraphs/UnivalentFamilies.lagda
IanRay11: I'm actually not sure because the edge that characterizes Magma's is not simply an equivalence between underlying types, but also states the equiv is structure preserving. Maybe I misunderstand.

source/ReflexiveGraphs/UnivalentFamilies.lagda
IanRay11: Good point. Adding now.

source/ReflexiveGraphs/UnivalentFamilies.lagda
IanRay11: Yes I will add a pointer to the index.

source/ReflexiveGraphs/UnivalentFamilies.lagda
tomdjong: Ah, I see, I was too quick, sorry. So what is this machinery intended to capture then? 
I guess any X with an embedding j : X -> Y would certainly work, right?

IanRay11: As the examples indicate any sub-universe like structure that is characterized by equivalences at the fibers of the map to Type. Of course h-level sub-universes offer a wealth of examples but the (ℕ , finset) example is interesting. 

Also, admittedly I haven't used this idea extensively beyond the (𝓤 , id) example (which will show up soon in my examples file). But I still think it may prove useful!

IanRay11: @jonsterling can you weigh in if you have a moment?

jonsterling: I'm not exactly sure what the question is here; this machinery does what it says on the tin — it defines univalent families. By definition, such things are always subuniverses of the meta-universe of all types — because that's what the univalence condition does.

tomdjong: Makes sense. Thanks, Jon!

IanRay11:
@tomdjong I see that you resolved the open issue. So this is ready for another round of review

tomdjong:
@martinescardo can take the last look before merging, if he wants to.

martinescardo:
I'll take a look later today.

source/ReflexiveGraphs/UnivalentFamilies.lagda
```diff
@@ -0,0 +1,88 @@
...
+  II x = ≃-refl (B x)
+
+is-univalent-family : Σ A ꞉ 𝓤 ̇ , (A → 𝓣 ̇)
+                    → 𝓤 ⊔ 𝓣 ̇
+is-univalent-family (A , B) = is-univalent-refl-graph (refl-graph-image A B)
+
+\end{code}
+
+This definition is intentionally general and offers a robust theoretical
+framework for working with a universe of things. That is, if one was working
```
martinescardo: Can we say something more precise than "things"?

martinescardo (approved):
It all looks good to me, with one minor "thing". :-)

martinescardo (approved):
I am happy with this now.

PR #433, opened 2026-04-09
Add a more direct correctness proof

cj-xu:


source/EffectfulForcing/Internal/AnotherCorrectnessProof.lagda
```diff
@@ -0,0 +1,346 @@
...
+  dialogue (η z)   α = z
+  dialogue (β φ x) α = dialogue (φ(α x)) α
+
+We define corresponding leaf and branching algebras in System T,
+thereby internalizing the dialogue function itself.
+
+\begin{code}
+
+⌜leaf⌝ : {Γ : Cxt}
+   → T Γ (ι ⇒ type-2)
```
martinescardo: Please indent all arrows at the same level as colons to conform to our formatting guidelines (I don't know if this is explicitly said - I should revise them eventually when I have time - but we do this in practice).

martinescardo: And, if there is not enough space to fit this in a line of length 80, the please do something like
```
blahhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh :
 : A
 → B
 → C
blahhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh 
 <parameters>
 = <something>
 where
  <whatever>
```

source/EffectfulForcing/Internal/AnotherCorrectnessProof.lagda
```diff
@@ -0,0 +1,346 @@
...
+  ξ,,θ (∈Cxt0 _) = θ
+  ξ,,θ (∈CxtS _ i) = ξ i
+FundamentalLemma α (t · u) γ δ ξ =
+   FundamentalLemma α t γ δ ξ
+      (⟦ u ⟧ γ) (⟦ ⌜ u ⌝ ⟧ δ)
+      (FundamentalLemma α u γ δ ξ)
+
+\end{code}
+
+Correctness theorem
```
martinescardo: Add full stop here.

martinescardo (commented):
This is very nice. I only had minor stylistic suggestions based on what we do in practice in TypeTopology.

But now I ask you one more thing. Please add a reference, at the top, to your own work on which these ideas are based.

Your proof is very nice, much better than the one we gave originally (which is still available here).

martinescardo (commented):
Great!

PR #437, opened 2026-04-29
Internal dialogue-tree correctness in PaperIndex

cj-xu:
Theorem 37 is about the correctness of the internal dialogue-tree function.


martinescardo (approved):
Thanks a lot!

ayberkt:
Thanks @cj-xu!

PR #445, opened 2026-06-18
Commutative idempotent binary operations with eventually idempotent right action imply sethood

fredrik-bakke:
Gives a generalisation of David Wärn's result that the algebraic structure of a semilattice implies sethood.

**Theorem.** Given a type $A$ with a binary operation ${-}*{-}$ that satisfies
1. idempotence ($x * x = x$)
2. commutativity ($x * y = y * x$)
3. eventual idempotence of the right action, meaning that for each $y$ there is a natural number $n$ such that, letting $r(x) := x * y$, then $r^{n+1}(x) = r^n(x)$ for all $x$

then $A$ is a set.


source/AlgebraicStructuresForcingSethood/CommutativeLoopSpaces.lagda
```diff
@@ -73,20 +73,15 @@ module idempotent-commutative-operation
...
-                    (＊Ω-commutative commutative refl q)
-      II₂ = ＊Ω-interchange-∙ refl p q refl
-      II₃ = ap₂ _＊Ω_ refl-left-neutral (refl-right-neutral p)
-      II₄ = ＊Ω-commutative commutative q p
-      II₅ = ap₂ _＊Ω_ (refl-left-neutral ⁻¹) ((refl-right-neutral q) ⁻¹)
-      II₆ = (＊Ω-interchange-∙ refl q p refl) ⁻¹
+      II₁ = ＊Ω-interchange-∙ p refl refl q
+      II₂ = ap₂ (_＊Ω_) (refl-right-neutral p) (refl-left-neutral)
+      II₃ = ap₂ (_＊Ω_) (refl-left-neutral ⁻¹) (refl-right-neutral q)
+      II₄ = ＊Ω-interchange-∙ refl q p refl ⁻¹
```
fredrik-bakke, line 84: I noticed that this proof could be simplified to avoid commutativity of `＊Ω`, but I know you prefer to preserve old formalisation. Should I add the simplified proof at the bottom of this file as a separate addition instead?

fredrik-bakke: @tomdjong

fredrik-bakke:
I'll mark this PR as ready for review once I've found the time to re-review the comments I received on my CSB-WLPO pull request, to avoid having you repeat feedback you've given me before.

fredrik-bakke:
~~Actually, shortly after posting this I realised I have a couple more generalisations of these facts.~~

EDIT: Nevermind, I've just stared at this for too long and started imagining things.

source/AlgebraicStructuresForcingSethood/CommutativeLoopSpaces.lagda
tomdjong: Nice! This seems like a small enough change (and a simplification too) that you can just edit the file. Please add a short comment at the top of the file, e.g. "Minor simplification by Fredrik Bakke on 18 June 2026".

tomdjong:
I'm going to unsubscribe from this PR until it's ready for PR (just to avoid a lot of emails about commits). Please ping me when it's ready. Thanks!

martinescardo:
> Gives a generalisation of David Wärn's result that the algebraic structure of a semilattice implies sethood.
> 
> **Theorem.** Given a type A with a binary operation − ∗ − that satisfies
> 
>     1. idempotence (
>          x
>          ∗
>          x
>          =
>          x
>        )
> 
>     2. commutativity (
>          x
>          ∗
>          y
>          =
>          y
>          ∗
>          x
>        )
> 
>     3. eventual idempotence of the right action, meaning that for each 
>          y
>         there is a natural number 
>          n
>         such that, letting 
>          r
>          (
>          x
>          )
>          :=
>          x
>          ∗
>          y
>        , then 
>          
>            r
>            
>              n
>              +
>              1
>            
>          
>          (
>          x
>          )
>          =
>          
>            r
>            n
>          
>          (
>          x
>          )
>         for all 
>          x
> 
> 
> then A is a set.

Is this motivated by something you have in mind?

source/UF/Sets.lagda
```diff
@@ -52,6 +52,11 @@ refl-is-set : (X : 𝓤 ̇ )
             → is-set X
 refl-is-set X r {x} p refl = r x p
 
+refl-is-set' : (X : 𝓤 ̇ )
```
martinescardo, line 55: Better name: `if-every-loop-is-refl-then-the-type-is-a-set`.

fredrik-bakke:
> Is this motivated by something you have in mind?

The motivation was to see how far I could stretch the original result. I have some new results today though, so better conditions will/might follow soon.

source/UF/Sets.lagda
fredrik-bakke: Sorry, does that mean I should rename `refl-is-set` too?

source/AlgebraicStructuresForcingSethood/CommutativeLoopSpaces.lagda
fredrik-bakke: I don't think this simplification is that minor, because it removes the only dependence on full commutativity of `_＊Ω_`. Other dependencies only need `refl` to be central.

PR #447, opened 2026-06-19
The loop space of the circle is the integers

tomdjong:
This PR also adds a few other useful bits, like the universal property of singletons.
Note that I work with a circle with definitional computation rules, implemented via postulates and rewrite rules, so that this bit of the development lives in the unsafe fragment of TypeTopology.

### DIFF
--- a/source/SyntheticHomotopyTheory/Circle/WithRewriting.lagda
+++ b/source/SyntheticHomotopyTheory/Circle/WithRewriting.lagda
@@ -60,7 +60,7 @@ so we replace [ap f loop] by its unfolding to a transport.
 
 \end{code}
 
-The above rewrite rules amount to the following defintional equalities.
+The above rewrite rules amount to the following definitional equalities.
 
 \begin{code}
 
--- a/source/SyntheticHomotopyTheory/Circle/WithRewriting.lagda
+++ b/source/SyntheticHomotopyTheory/Circle/WithRewriting.lagda
@@ -1,4 +1,4 @@
-Tom de Jong, 18 & 19 June 2026.
+Tom de Jong, 19 June 2026.
 
 We postulate the existence of the circle S¹ with definitional computation rules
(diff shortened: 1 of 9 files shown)


source/SyntheticHomotopyTheory/Circle/WithRewriting.lagda
```diff
@@ -0,0 +1,162 @@
...
+                      → S¹-recursion A a l pt ＝ a
+
+ {-# REWRITE S¹-recursion-comp-pt #-}
+
+\end{code}
+
+In the following computation rule, we would like to write
+   ap f loop ＝ l
+but by definition the left-hand side reduces to a transport,
+and Agda only allows rewriting when the left-hans side does not reduce,
```
martinescardo, line 38: Typo `left-hans`.

source/UF/Singleton-Properties.lagda
```diff
@@ -68,3 +69,64 @@ the-singletons-form-a-singleton-type {𝓤} fe pe =
               (λ (j , x) → pointed-props-are-singletons x j)
 
 \end{code}
+
+Added 19 June 2026 by Tom de Jong.
+
+\begin{code}
+
+consts : (A : 𝓤 ̇ ) (X : 𝓥 ̇ ) → X → (A → X)
```
martinescardo, line 77: Why `consts` rather than `const`? Just curious.

martinescardo (approved):
Great. There is a minor point. You can merge yourself when you are ready.

source/UF/Singleton-Properties.lagda
tomdjong: I refer to the map as "the constants map". (Calling it "the constant map" would be confusing because the map itself isn't constant; it just _returns_ constant maps.)

tomdjong:
Thanks for the very quick review!

PR #453, opened 2026-07-13
Construct infima of inhabited families of ordinals

nicolaikraus:
This new file gives several constructions of infima of small families of ordinals. It's essentially the dual of Tom's construction of suprema.
The construction is very simple and arguably unsurprising, but it is (in my opinion) a natural addition to the Ordinals development.

Note: I realise that there is a name clash. The file Ordinals.InfProperty also talks about (a different kind of) infima.


source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+in the original family.
+
+The construction is this:
+
+    Inf F := Σ f ꞉ ((i : I) → ⟨ F i ⟩) , (∀ j k → F j ↓ f j ≃ₒ F k ↓ f k),
+
+ordered pointwise by (f , _) ≺ (g , _) := ∀ i → f i ≺ g i. This file proves that
+Inf F is indeed the infimum.
+
+Note: Tom de Jong's construction of suprema (in Ordinals.OrdinalOfOrdinalsSuprema)
```
martinescardo: There are actually two constructions of suprema, one by Tom and one by me. Tom implemented both, after I told him my construction in my whiteboard.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+requires set-quotients. The current construction of Infima does not require
+these; apart from univalence, only propositional truncation is required to
+express that I is inhabited.
+
+Christian Sattler pointed out to me that there is also another simple construction:
+The infimum is the supremum of all ordinals that are below everything in the family.
+
+NOTE: There is also a different notion of Infimum by Escardo in the file
+Ordinals.InfProperty.lagda from 2012.
```
martinescardo: May be worth saying that this is an infimum within an ordinal, rather than within the ordinal of ordinals. 

But actually this doesn't matter much, because any ordinal is embedded in the ordinal of ordinals as an initial segment.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+ fe' {𝓤} {𝓥} = fe 𝓤 𝓥
+
+open import Ordinals.Arithmetic fe
+open import Ordinals.AdditionProperties ua
+open import Ordinals.OrdinalOfOrdinalsSuprema ua
+open import UF.Size
+
+\end{code}
+
+Some basic lemmas that help proving that a type with relation is an ordinal.
```
martinescardo: "type with relation" -> "type equipped with a relation", maybe?

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+
+\begin{code}
+
+Ordinal-Of-Ordinals-Has-Small-Infima : (𝓤 : Universe) → 𝓤 ⁺ ̇
+Ordinal-Of-Ordinals-Has-Small-Infima 𝓤 =
+   (I : 𝓤 ̇ ) → ∥ I ∥ → (F : I → Ordinal 𝓤)
+ → Σ γ ꞉ Ordinal 𝓤 , ((i : I) → γ ⊴ F i)
+                   × ((β : Ordinal 𝓤) → ((i : I) → β ⊴ F i) → β ⊴ γ)
+
+Ordinal-Of-Ordinals-Has-Small-Infima-is-prop :
```
martinescardo, line 160: This departs a little bit from our notational conventions. Firstly, we should use lower case for this. Secondly, we like "being". So e.g.

"having-small-infima-is-prop". 

Or you can make it longer if you wish.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+module inf-construction
+        {𝓤 : Universe}
+        {I : 𝓤 ̇ }
+        (F : I → Ordinal 𝓤)
+       where
+
+ is-compatible : ((i : I) → ⟨ F i ⟩) → 𝓤 ̇
+ is-compatible f = (j k : I) → (F j ↓ f j) ≃ₒ (F k ↓ f k)
+
+ is-compatible-is-prop : (f : (i : I) → ⟨ F i ⟩) → is-prop (is-compatible f)
```
martinescardo: `being-compatible-is-prop`, again to conform to our conventions.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+          (Inf-Ord δ ↓ from-gcis w)      ∎)
+    where
+     step = (simulations-preserve-↓ (Inf-Ord δ) gcis inf-⊴-gcis (from-gcis w)) ⁻¹
+
+ gcis-is-infimum : gcis ＝ inf δ F
+ gcis-is-infimum = ⊴-antisym gcis (inf δ F) gcis-⊴-inf inf-⊴-gcis
+
+\end{code}
+
+Another simple observation: infima commute with imfima. An infimum over a Σ-type is
```
martinescardo: Typo: "imfima".

martinescardo (changes requested):
Interesting!

I have some minor things for you to fix.

@tomdjong Could I ask you to do a review as well? Thanks.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
tomdjong: Moreover, the construction attributed to me here was taken from the HoTT Book (Lemma 10.3.22). I just observed that it was also the _least_ upper bound.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+
+Note: Tom de Jong's construction of suprema (in Ordinals.OrdinalOfOrdinalsSuprema)
+requires set-quotients. The current construction of Infima does not require
+these; apart from univalence, only propositional truncation is required to
+express that I is inhabited.
+
+Christian Sattler pointed out to me that there is also another simple construction:
+The infimum is the supremum of all ordinals that are below everything in the family.
+
+NOTE: There is also a different notion of Infimum by Escardo in the file
```
tomdjong: Why are _Infimum_ and _Infima_ getting capitalized? They're just ordinary words. I would also lowercase `Inf F`.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+
+ordered pointwise by (f , _) ≺ (g , _) := ∀ i → f i ≺ g i. This file proves that
+Inf F is indeed the infimum.
+
+Note: Tom de Jong's construction of suprema (in Ordinals.OrdinalOfOrdinalsSuprema)
+requires set-quotients. The current construction of Infima does not require
+these; apart from univalence, only propositional truncation is required to
+express that I is inhabited.
+
+Christian Sattler pointed out to me that there is also another simple construction:
```
tomdjong: In a complete lattice, this fact is well known, and of course, the converse holds too. If we have infima, then we have suprema by taking the greatest lower bound of all upper bounds. What is noteworthy here is that a priori this won't work because we only have suprema (and/or infima) of _small_ families.

I suspect (but haven't thought it about it much)  that you can't reverse the implications, i.e. construct infima as you do and then construct suprema (without set quotients!) from those. Or can you?

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+       where
+
+open import MLTT.Spartan
+open import Ordinals.Equivalence
+open import Ordinals.Maps
+open import Ordinals.Notions
+open import Ordinals.OrdinalOfOrdinals ua
+open import Ordinals.Type
+open import Ordinals.Underlying
+open import UF.Base
```
tomdjong:
```
om@pop-os ~/r/n/source (infima)> ../imports.sh Ordinals/OrdinalOfOrdinalsInfima.lagda 
Importing UF.Base was not necessary
Importing UF.Equiv was not necessary
Importing UF.LeftCancellable was not necessary
```
Indeed, these imports are not needed.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+        seg = pr₁ ([ β , F i₀ ]⟨ h i₀ ⟩-is-simulation) b (f i₀) (l i₀)
+
+\end{code}
+
+We assumed that I is inhabited. What happens if we drop the assumption? If I is
+empty, then Inf F is not an ordinal because the order as defined above is reflexive,
+vacuously. (There cannot be a greatest lower bound of the empty set, because it
+would have to be an upper bound for all ordinals, which is impossible by
+Burali-Forti.) If Inf F is an ordinal, we show ¬¬I in the next lemma.
+Much further below, we show that the un-negated conclusion is a constructive taboo.
```
tomdjong: _unnegated_ (?)

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+       where
+
+ is-compatible : ((i : I) → ⟨ F i ⟩) → 𝓤 ̇
+ is-compatible f = (j k : I) → (F j ↓ f j) ≃ₒ (F k ↓ f k)
+
+ is-compatible-is-prop : (f : (i : I) → ⟨ F i ⟩) → is-prop (is-compatible f)
+ is-compatible-is-prop f =
+  Π₂-is-prop fe' (λ j k → ≃ₒ-is-prop-valued fe' (F j ↓ f j) (F k ↓ f k))
+
+ Inf : 𝓤 ̇
```
tomdjong: lowercase

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+ (f , _) ≺ (g , _) = (i : I) → f i ≺⟨ F i ⟩ g i
+
+ ≺-is-prop-valued : is-prop-valued _≺_
+ ≺-is-prop-valued (f , _) (g , _) =
+  Π-is-prop fe' (λ i → Prop-valuedness (F i) (f i) (g i))
+
+\end{code}
+
+For every i : I the "projection" c : Inf → F i is monotone, an embedding, and
+satisfies the simulation property. (None of this requires I to be inhabited.)
```
tomdjong: The comment in parentheses does not really make sense to me, because the statements start with the assumption _for every i : I_, so you get to assume that _I_ is inhabited for free.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+
+\begin{code}
+
+ module _ (i : I) where
+
+  c : Inf → ⟨ F i ⟩
```
tomdjong: I think these projections should either have a better name (I think that would be preferred option, since you might wish to refer to these outside this file), or they should be `private`.

EDIT: I see these there is an exporting module later, so probably these should just be made private.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+    aₑ = (a , l)
+    w : (j : I) → ⟨ F j ↓ g j ⟩
+    w j = ≃ₒ-to-fun (F i ↓ g i) (F j ↓ g j) (q i j) aₑ
+    h : (j : I) → ⟨ F j ⟩
+    h j = pr₁ (w j)
+    hb : (j : I) → h j ≺⟨ F j ⟩ g j
+    hb j = pr₂ (w j)
+    key : (j : I) → (F j ↓ h j) ＝ (F i ↓ a)
+    key j =
+     (F j ↓ h j)
```
tomdjong: We usually display this as
```
X ＝⟨ I   ⟩
Y ＝⟨ II  ⟩
Z ＝⟨ III ⟩
 where
  I   = ...
  II  = ...
  III = ...
```
Since `I` is taken in this context, you could use `⦅1⦆`, `⦅2⦆`,... or `[1]`, `[2]`,...

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+     ((F i ↓ g i) ↓ aₑ)
+      ＝⟨ iterated-↓ (F i) (g i) a l ⟩
+     (F i ↓ a) ∎
+    r : is-compatible h
+    r j k = idtoeqₒ (F j ↓ h j) (F k ↓ h k) (key j ∙ (key k) ⁻¹)
+
+\end{code}
+
+By the lemma from the beginning, using any i₀ : I extracted from the assumption
+that I is inhabited, ≺ is a well-order, so Inf is an ordinal Inf-Ord : Ordinal 𝓤.
```
tomdjong: Nice!

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+
+We assumed that I is inhabited. What happens if we drop the assumption? If I is
+empty, then Inf F is not an ordinal because the order as defined above is reflexive,
+vacuously. (There cannot be a greatest lower bound of the empty set, because it
+would have to be an upper bound for all ordinals, which is impossible by
+Burali-Forti.) If Inf F is an ordinal, we show ¬¬I in the next lemma.
+Much further below, we show that the un-negated conclusion is a constructive taboo.
+
+\begin{code}
+ well-founded-gives-¬¬-inhabited : is-well-founded _≺_ → ¬¬ I
```
tomdjong, line 343: Add a blank line.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
tomdjong: I think I also used upper case in the suprema file to distinguish it from the _proof_ (term) that Ord has small suprema, and the same applies here to infima.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+
+ inf-reindex : {I J : 𝓤 ̇ } (ε : ∥ J ∥) (ρ : J → I) (F : I → Ordinal 𝓤)
+             → inf (∥∥-functor ρ ε) F ⊴ inf ε (F ∘ ρ)
+ inf-reindex ε ρ F =
+  inf-is-greatest-lower-bound ε (F ∘ ρ) (inf (∥∥-functor ρ ε) F)
+   (λ j → inf-is-lower-bound (∥∥-functor ρ ε) F (ρ j))
+
+\end{code}
+
+The "dual" or the standard result for suprema:
```
tomdjong: _of_ the standard

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+        {𝓤 : Universe} {I : 𝓤 ̇ }
+        (δ : ∥ I ∥) (F : I → Ordinal 𝓤) (i₀ : I)
+       where
+
+ open inf-construction F
+
+ is-common-lower-segment : ⟨ F i₀ ⟩ → 𝓤 ̇
+ is-common-lower-segment x = (j : I) → Σ y ꞉ ⟨ F j ⟩ , (F i₀ ↓ x) ≃ₒ (F j ↓ y)
+
+ is-common-lower-segment-is-prop : (x : ⟨ F i₀ ⟩)
```
tomdjong: Martín's _being_ comment applies here too.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+ is-common-lower-segment-is-prop x = Π-is-prop fe' γ
+  where
+   γ : (j : I) → is-prop (Σ y ꞉ ⟨ F j ⟩ , (F i₀ ↓ x) ≃ₒ (F j ↓ y))
+   γ j (y , e) (y' , e') =
+    to-subtype-＝ (λ y → ≃ₒ-is-prop-valued fe' (F i₀ ↓ x) (F j ↓ y))
+     (↓-lc-bis (F j) y y'
+       (≃ₒ-trans (F j ↓ y) (F i₀ ↓ x) (F j ↓ y')
+         (≃ₒ-sym (F i₀ ↓ x) (F j ↓ y) e) e'))
+
+ is-common-lower-segment-is-lower-closed :
```
tomdjong: And here too, I think.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
tomdjong: Or like you do on line 597.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
```diff
@@ -0,0 +1,711 @@
...
+        subtype-order-is-prop-valued  (F i₀) is-common-lower-segment ,
+        subtype-order-is-well-founded (F i₀) is-common-lower-segment ,
+        ≪-is-extensional ,
+        subtype-order-is-transitive   (F i₀) is-common-lower-segment
+
+\end{code}
+
+The projection (f , p) ↦ f i₀ corestricts to a map Inf → GCIS, which is a
+simulation; the map that reads off, for each j, the point realising the common
+segment is a section of it. That makes it an order-isomorphism, so gcis is the
```
tomdjong: Grammar issue / missing word?

tomdjong (commented):
Cool!

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
nicolaikraus: Thanks, I missed two references.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
nicolaikraus: I think it matters because having infima within an ordinal is, in general, stronger than having infima once the ordinal is embedded into the ordinal of ordinals. Concretely:
Let A be an ordinal and Ord be the ordinal of ordinals. There is _! : A -> Ord. Suppose I is a small set and F : I -> carrier(A) a function. In general, we can't construct an infimum of F in A, but we can construct the infimum of _! \circ F in Ord.
Anyway, I'll add what you suggested.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
nicolaikraus: yes, that sounds better.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
nicolaikraus: Yes, what I wrote there is a direct copy of OrdinalOfOrdinalsSuprema.lagda. I would prefer both to use the same convention because they're mostly dual.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
nicolaikraus: thanks, I'll fix it.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
nicolaikraus: thanks

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
nicolaikraus: You're right, it should be "infimum" and `inf F`

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
nicolaikraus: I agree. Not sure what I was thinking there.

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
nicolaikraus: Ok!

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
nicolaikraus: I don't see the mistake.

tomdjong: I think that the additional phrase ", for each j," was enough for me to have trouble parsing this sentence.
Maybe it's easier like this: _this map has a section which, for each j, reads off the point realising the common segment_?

source/Ordinals/OrdinalOfOrdinalsInfima.lagda
nicolaikraus: Right, you can't do this because there's no small replacement.

tomdjong: Maybe it's worth writing a comment explaning this. But I'll leave the decision up to you.

nicolaikraus:
I've implemented all fixes. (edit: and fixed typos in the last commit.)

tomdjong:
@martinescardo Shall I merge this?

martinescardo:
Sorry for the delay. I was on holiday.

tomdjong:
No worries. Hope you had a nice time! :)

nicolaikraus:
Thanks!

